PR #8, opened 2016-03-10
Tw date

adrienDog:
Support for:
- locale or ngLocale
- min / ngMin
- max/ngMax

Corresponding validation classes should apply

styleguide_components_-_transferwise


adrienDog:
@StevePole I have changed `ngLocale` to `twLocale` :)

Issue #11, opened 2016-03-29
International month names not upper case in FF


StevePole:
@adrienDog could you address this at some point, I believe it's in git history.

PR #12, opened 2016-03-29
Fix month names + Add Firefox in karma

adrienDog:


adrienDog:
To fix issue #11 @StevePole

angular/components/tw-date/tw-date.controller.js
```diff
@@ -314,7 +310,7 @@
 		}
 
 		function getIsoDateWithoutTime(dateAsISOString) {
-			return dateAsISOString.substr(0, dateAsISOString.indexOf('T'));
+			return dateAsISOString.substring(0, dateAsISOString.indexOf('T'));
```
StevePole: Why this change?

angular/components/tw-date/tw-date.directive.spec.js
```diff
@@ -473,6 +473,7 @@ describe('Directive: TwDate', function() {
             });
         });
         describe('with month select', function() {
+            // TODO: Fix for Firefox
```
StevePole: TODO?

angular/components/tw-date/tw-date.controller.js
adrienDog: just for consistency with the other usages and also because I felt the definition of `substring` was more similar to what we want here (even though in practice it is the same effect here)
http://stackoverflow.com/questions/3745515/what-is-the-difference-between-substr-and-substring

angular/components/tw-date/tw-date.directive.spec.js
adrienDog: It appeared when I added FF as a browser in karma: it works in real FF, but in the tests somehow I think the triggerHandler doesn't work.
I'll take a quick look sry didnt think it was urgent to tackle

StevePole: It might not be urgent, I just saw "Fix for Firefox" and thought, surely we want to support firefox...  Perhaps explain that it does not affect the real browser.

adrienDog: It does sound scary when you read it ^^

StevePole:
:+1:

PR #26, opened 2016-09-21
Currency input

igalarzab:


StevePole:
Thanks, I will do a full review, but first comment: TESTS!  There should be some good examples to follow within the other components.

igalarzab:
Sorry Steve, it's not finished yet. I create the PR for myself. I'll ping you when I finish (there will be tests, don't worry)

igalarzab:
There is something that doesn't make sense to me. Having separate `currency-code` and `currency-symbol` parameters in the component. If this is not a generic widget but one to use just for currencies, one parameter should be enough.

@StevePole What do you think about linking this library with [currencies](https://www.npmjs.com/package/currencies) and requiring just the code?

StevePole:
Hmmm, yes, I quite like the idea of just passing currency code, however, I'm also trying to limit external dependencies as well.  I would rather have a self-contained mapping of code => symbol, there are only a finite number of currencies after all.  I just wonder whether it's easier to go ahead with the symbol/code approach?  Also, not sure about localisation of codes, if we're in a cyrillic language maybe we would pass the localised code rather than the latin characters?

angular/components/tw-currency-input/tw-currency-input.directive.js
```diff
@@ -0,0 +1,69 @@
...
+				return true;
+			}
+
+			return viewValue <= scope.vm.ngMax;
+		};
+	}
+
+	var templateAsString = ' \
+		<div class="input-group"> \
+			<span class="input-group-addon input-currency-symbol">{{ vm.currencySymbol }}</span> \
```
StevePole: The convention I've been using is to use classes related to the component name e.g. `.tw-currency-input-symbol, .tw-currency-input-code` to 'namespace' these classes that are largely there to decouple the markup structure from the tests.

angular/components/tw-currency-input/tw-currency-input.directive.spec.js
```diff
@@ -0,0 +1,159 @@
...
+        });
+
+        it('should set ngModel.$touched when blured', function() {
+            input.triggerHandler('blur')
+            expect($ngModel.$touched).toBe(true);
+        });
+
+        it('should style nearest parent form-group when focussed', function() {
+            input.triggerHandler('focus');
+            expect(directiveElement.closest('.form-group').hasClass('focus')).toBe(true);
```
StevePole: Surprised this works, I don't see tw-focusable on the input...

partials/tw-currency-input.html
```diff
@@ -0,0 +1,95 @@
...
+    ng-required="{{ vm.currencyInput.required }}"
+    ng-min="{{ vm.currencyInput.min }}"
+    ng-max="{{ vm.currencyInput.max }}" /&gt;</pre>
+    </div>
+    <div class="col-md-6">
+        <div class="well">
+            <h5 class="page-header">
+                Edit configuration
+            </h5>
+			<div class="form-group">
```
StevePole: Looks like some whitespace issue in this file.

angular/components/tw-currency-input/tw-currency-input.controller.js
```diff
@@ -0,0 +1,34 @@
...
+			}
+		});
+	}
+
+	function validateInput($ngModel, $element) {
+		if (!$ngModel.$touched) {
+			return;
+		}
+
+		if ($ngModel.$invalid) {
```
StevePole: Can you remove this if you add `[tw-validation]` to the component declaration?  Ideally, validation is opt-in.  If the ng-model of the component is invalid it should do this based on that directive.

angular/components/tw-currency-input/tw-currency-input.directive.js
```diff
@@ -0,0 +1,69 @@
...
+			return viewValue >= scope.vm.ngMin;
+		};
+
+		ngModel.$validators.max = function(modelValue, viewValue) {
+			if (typeof scope.vm.ngMax === "undefined" || !viewValue) {
+				return true;
+			}
+
+			return viewValue <= scope.vm.ngMax;
+		};
```
StevePole: I know you copied this structure from somewhere else, so it's my fault :-), but you could do this in the controller, it will be more future proof if we turn these into 1.5 components as they have no link function.  `element` here is the same as the `$element` you inject into your controller, `ngModel` is the same controller you got with `var $ngModel = $element.controller('ngModel');`

GruntFile.js
```diff
@@ -5,11 +5,11 @@ module.exports = function(grunt) {
         watch: {
             less: {
                 files: ['styles/**/*.less'],
-                tasks: ['less']
+                tasks: ['less', 'copy']
             },
             components: {
                 files: ['angular/**/*.js'],
-                tasks: ['uglify', 'jshint']
+                tasks: ['uglify', 'jshint', 'copy']
```
StevePole: Why did you need this change?

igalarzab: Forget it, it's not necessary. I was testing my changes inside of the gh-pages folder instead of in the root dir

igalarzab:
Thanks for the review @StevePole.

About the currencies list, I'm afraid of creating a different list each time we need it in different projects (it creates inconsistency problems). What are your concerns about adding a third party lib? Does it has something to do with [this](http://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos/)? We can create our own package (but I wonder if maybe it's complicating too much this thing).

PR #27, opened 2016-09-23
Add change behaviour to the currency input

igalarzab:


angular/components/tw-currency-input/tw-currency-input.controller.js
```diff
@@ -8,6 +8,7 @@
 	TwCurrencyInputController.$inject = ['$element', '$scope'];
 
 	function TwCurrencyInputController($element, $scope) {
+		var vm = this;
```
StevePole: Prefer $ctrl now, more compatible with 1.5 components.

PR #28, opened 2016-09-26
minValue null and $timeout

igalarzab:


angular/components/tw-currency-input/tw-currency-input.controller.js
```diff
@@ -23,15 +23,15 @@
 		});
 
 		$ngModel.$validators.min = function(modelValue, viewValue) {
-			if (typeof $scope.vm.ngMin === 'undefined' || viewValue === null) {
+			if (typeof $scope.vm.ngMin === 'undefined' || $scope.vm.ngMin === null || !isNumber(viewValue)) {
```
StevePole: Why does this use scope, can't you use $ctrl?

angular/components/tw-currency-input/tw-currency-input.controller.js
```diff
@@ -40,9 +40,15 @@
...
 		$ctrl.changedInputValue = function() {
 			if ($ctrl.ngChange) {
-				$ctrl.ngChange();
+				// $timeout is needed to get the last ngModel value.
+				// See: https://github.com/angular/angular.js/issues/4558
+				$timeout($ctrl.ngChange);
 			}
 		};
+
+		function isNumber(value) {
```
StevePole: angular has a built in function angular.isNumber https://docs.angularjs.org/api/ng/function/angular.isNumber which might be more concise?

angular/components/tw-currency-input/tw-currency-input.directive.js
```diff
@@ -30,12 +31,14 @@
 		<div class="input-group"> \
 			<span class="input-group-addon tw-currency-input-symbol">{{ vm.currencySymbol }}</span> \
 			<input \
-				type="number" \
+				type="tel" \
 				autocomplete="off" \
 				name="amount" \
 				step="any" \
 				class="form-control text-xs-right p-r-0" \
+				show-decimals="vm.showDecimals" \
```
StevePole: I'd prefer to use controllerAS: '$ctrl' rather than vm, because we won't have to change the template when we move it to a 1.5 component.

angular/components/tw-currency-input/tw-currency-input.directive.spec.js
```diff
@@ -64,7 +64,7 @@ describe('Directive: TwCurrencyInput', function() {
 
         it('should change ngModel when value is changed', function() {
             input.val('200.05').trigger('input');
-            expect($scope.ngModel).toBe(200.05);
+            expect($scope.ngModel).toBe('200.05');
```
StevePole: Why is ngModel a string?

igalarzab: Because it's not a `type="number"` anymore and I don't have any $parser to transform the value to an integer.

angular/components/tw-currency-input/tw-currency-input.controller.js
igalarzab: Uhm, I checked the documentation but seems that doesn't work for strings, so I would need to transform it to a Number anyway (or worse, do something like `+value` to force the cast 💩 )

StevePole: 👍

angular/components/tw-currency-input/tw-currency-input.directive.spec.js
StevePole: Ok, but what does the consumer of this component expect?

igalarzab: A string? :trollface: 









Ok, let me work on it, I have to check if it will collide with the `tw-number-input-formatter` directive

angular/components/tw-currency-input/tw-currency-input.directive.js
```diff
@@ -20,6 +20,7 @@
 				ngMin: '=',
 				ngMax: '=',
 				ngRequired: '=',
+				showDecimals: '=',
```
StevePole: Not sure we should let the consumer choose this?  Might be best to always show them, could have a quick chat to Patrick/Mike about it.

igalarzab: My thought was that there are currencies without decimals. There are also currencies with 3 decimals, but it seems that our API doesn't support them (the field it's named `hasDecimals`, not `numDecimal`)

StevePole: Thats sucks, could we change this to 'decimals' and expect a number, that way we're not infected by the errors above?

PR #31, opened 2016-10-10
Remove spacing between <i> and transcluded elements

theSKa:
If there is no title set on the droppable and the consumer wants to convey info using the transclusion then the m-t-2 spacing makes the content look odd.


StevePole:
👍

PR #33, opened 2016-10-11
Fully dynamic form and passwords in dynamic input

StevePole:


angular/components/tw-dynamic-form/tw-dynamic-form.directive.spec.js
```diff
@@ -0,0 +1,296 @@
...
+
+        describe('in tab navigation', function() {
+            it('shows a tab for every requirement', function() {
+                var navWrapper = directiveElement.find('.nav.nav-tabs');
+                var navTabs = navWrapper.find('li');
+                expect(navWrapper.length).toBe(1);
+                expect(navTabs.length).toBe(REQUIREMENTS.length);
+            });
+
+            it('shows one active tab based on the model', function() {
```
StevePole: We can be more specific about the behaviour we expect, e.g. "shows the first tab as active”.  Important because we later go onto test the content of the tab without first being specific about which tab we expected.

angular/components/tw-dynamic-form/tw-dynamic-form.directive.spec.js
```diff
@@ -0,0 +1,296 @@
...
+                var navWrapper = directiveElement.find('.nav.nav-tabs');
+                var activeTab = navWrapper.find('li.active');
+                expect(activeTab.length).toBe(1);
+
+                var tabText = activeTab.find('a').text();
+                expect(tabText.trim()).toBe('iban');
```
StevePole: I'd prefer not to test the _text_ 'iban', it would be better to test "expect(tabText.trim()).toBe(Requirements[1].type);” as the test is less brittle (we don’t really care about the content, we just care that the right data from the requirements is used).

I don't understand why it's iban in this case,are we showing the second tab by default?

StevePole: Just realised I did exactly what I said not to do in the tw-select tests...

PR #34, opened 2016-10-13
Add locale support for day/month ordering in tw-date. Build.

StevePole:
Fix bug with 0 values in tw-select.


angular/components/tw-select/tw-select.directive.spec.js
```diff
@@ -296,24 +336,26 @@ describe('Directive: TwSelect', function() {
...
-                $(firstLink).focus().triggerHandler(keypress('ArrowDown'));
+                var secondLink = directiveElement.find('.active a')[0];
+                $(secondLink).focus().triggerHandler(keypress('ArrowDown'));
 
                 // Dropdown.js not working in tests
-                var secondLink = directiveElement.find('.active a')[0];
-                //expect($scope.ngModel).toBe('2');
-                //expect($(secondLink).text.trim()).toBe('Two');
+                var thirdLink = directiveElement.find('.active a')[0];
+                //expect($scope.ngModel).toBe(OPTIONS[2].value);
```
vahurh: Are you working on a fix for dropdown.js or are these things gonna stay commented out?

dist/js/form-components.js
```diff
@@ -60,7 +60,7 @@ angular.module("tw.form-components", []);
             vm.dateDisabled = void 0 !== vm.ngDisabled ? vm.ngDisabled : void 0 !== vm.disabled;
         }
         function setDateLocale() {
-            vm.locale || (vm.locale = DEFAULT_LOCALE_EN);
+            vm.locale || (vm.locale = DEFAULT_LOCALE_EN), vm.locale.indexOf("US", vm.locale.length - 2) !== -1 ? vm.monthBeforeDay = !0 : vm.monthBeforeDay = !1;
```
vahurh: Is there a reason for dist files being in the git repo? Also - I've never seen an expression like that. Cool that it works.

angular/components/tw-select/tw-select.directive.spec.js
StevePole: Not working on it at the moment, it's not so much fixing dropdown.js as removing the dependency and reimplementing the functionality.  I'd like to to do it, but it's not at the top of my list.

dist/js/form-components.js
StevePole: It's because the project is imported by bower/npm, you want to have the final files in the repo.  The minifier that combines the files does some interesting stuff, even when not uglifying.

angular/components/tw-date/tw-date.directive.js
```diff
@@ -103,7 +97,20 @@
 
 	var templateAsString = " \
 			<div class='row'> \
-				<div class='col-sm-3'> \
+				<div class='col-sm-5 tw-date-month-column' ng-if='vm.monthBeforeDay'> \
```
StevePole: Ok, I think it's better to make use of ng-if because it means you can update the locale at run time, rather than just component compile time.  But I like the idea of not repeating the interior markup.  How about I make a 'monthSelectTemplate' and just use that in both sections?

PR #39, opened 2016-10-28
Reduce amount currency border to the input-group

StevePole:
Remove symbol from currency input.
Add size support to currency and amount currency.
Update docs.


kdelemme:
👍

PR #43, opened 2016-11-03
Add show-decimals option to tw-amount-currency-select

lapostoj:
I couldn't really find a way to test it from here as `tw-number-input-formatter` is not present it doesn't format anyway.
But it follows the same logic used in tw-currency-input so I expect it to be ok.
If you have any idea to test, let me know :) (like is it possible to import it in another project without having it released?)

### DIFF
--- a/index.html
+++ b/index.html
@@ -38,7 +38,6 @@
 				<div data-spy="affix" data-offset-top="0">
 					<h5>Jump to:</h5>
 					<ul class="list-unstyled">
-						<li><a href="#amount-currency-select">Tw Amount Currency Select</a></li>
 						<li><a href="#focusable">Tw Focusable</a></li>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -87,9 +87,9 @@ <h3 class="panel-title">Tw Loader</h3>
 					<div class="panel-body">
 						<p>A directive to add the HTML necessary for our
 							loading indicator.  Simply use:</p>
-<pre>
-&lt;tw-loader /&gt;
-</pre>
+						<pre>
+							&lt;tw-loader /&gt;
+						</pre>
```
StevePole: Can you put this back?  Because it's in a pre the browser will respect this whitespace which is undesirable.  Other than that, looks good 👍

lapostoj: Oh sure! Didn't know that!

PR #47, opened 2016-11-22
Add support for disabled option in tw-select. Build.

StevePole:


angular/components/tw-select/tw-select.directive.js
```diff
@@ -178,7 +181,14 @@
 		function buttonFocus() {
 			$element.triggerHandler('focus');
 		}
-		function optionClick(option) {
+		function optionClick(option, $event) {
+			if (option.disabled) {
+				if ($event.stopPropagation) {
+					$event.stopPropagation();
+				}
+				$event.cancelBubble = true;
```
oliverviljamaa: Why do we need this in addition to `stopPropagation`?

oliverviljamaa:
Noticed some weird behaviour with `tab` click – it doesn't skip the disabled one and as there's no visible focus state, it seems to just not do anything.

angular/components/tw-select/tw-select.directive.js
StevePole: As far as I'm aware, stopProgogation doesn't have universal browser support.

oliverviljamaa: Seems to work everywhere except IE < 9:
https://developer.mozilla.org/en/docs/Web/API/Event/stopPropagation
`cancelBubble` is deprecated according to MDN:
https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/cancelBubble

StevePole: Ok, I'll simplify.

PR #49, opened 2016-12-01
Add tw process

StevePole:


angular/components/tw-process/tw-process.directive.js
```diff
@@ -0,0 +1,111 @@
...
+	function TwProcessController($scope, $interval, $timeout) {
+		var $ctrl = this;
+		$ctrl.processing = $ctrl.state;
+		var promise;
+
+		// This allows us to cancel the interval when not needed.
+		$scope.$watch('$ctrl.state', function(newVal) {
+			if ($ctrl.processing === -1 ||
+				$ctrl.processing === 0 ||
+				$ctrl.processing === 1) {
```
adrienDog: I have seen this condition several times (`$watch`, `startProcess`)
```
if ($ctrl.processing === -1 || $ctrl.processing === 0 || $ctrl.processing === 1)
```
What are we checking?

angular/components/tw-process/tw-process.directive.js
```diff
@@ -0,0 +1,111 @@
...
+				// We're stopped so restart
+				$ctrl.processing = null;
+				$ctrl.startProcess();
+			}
+		});
+
+		$scope.$watch('$ctrl.size', function(newVal) {
```
adrienDog: Do you foresee clients wanting a dynamic `size` parameter? Just wondering if this watcher is actually needed. My feeling is that clients know the size of this component is static, like a class for sizing would be

StevePole: Yes, I agree in most cases.  I actually did it because the docs break otherwise, but it is also nicer as clients can use it however they want.

adrienDog: I suspected the docs to be part of the decision :) but yeah more freedom is good I guess 👍

PR #54, opened 2016-12-14
CON-647 - Removes duplicates from search if multiple categories have …

indrekru:
…one currency


StevePole:
Could you add a test?

PR #67, opened 2017-03-13
Propagating the event from upload as well as the file

coolsebz:
Changes:
- in the old upload directive, select and drag now propagate the event too
- in the new upload directive: select and drag now propagate the event in `onStart`

Main uses for this change:
In batch payments we would like to track what method do our customers use for uploading, therefore being able to use the `event.type` property becomes useful

used in: https://github.com/transferwise/mass-pay-frontend/pull/93


ttqin:
LGTM

angular/components/tw-upload-droppable/tw-upload-droppable.directive.js
```diff
@@ -44,15 +44,15 @@
 		$ctrl.dragCounter = 0;
 		$ctrl.isActive = false;
 
-		$ctrl.onManualUpload = function() {
+		$ctrl.onManualUpload = function(evt) {
```
StevePole: Can you remove the contraction?  i.e. evt => event.  Otherwise all good 👍

PR #73, opened 2017-04-04
tw-date bug fix

StevePole:


StevePole:
@lapostoj could you take a look?

PR #76, opened 2017-04-19
Fix requirements form

lapostoj:


README.md
```diff
@@ -114,5 +119,21 @@ In order to release a new version, here are the steps necessary:
...
+        git checkout gh-pages
+    ```
+    Then, for every update: (checking that you are on the branch `gh-pages`)
+    ```
+        grunt (in the root of styleguide-components)
+        cd gh-pages
+        git add .
+        git commit -m "Update docs"
+        git push origin gh-pages
+    ```
```
StevePole: Thanks for adding 👍

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
```diff
@@ -126,11 +127,16 @@
...
-						button-text='Upload' \
-						accept='' \
+						icon='{{$ctrl.uploadOptions.icon}}' \
+						instructions='{{$ctrl.uploadOptions.instructions}}' \
+						button-text='{{$ctrl.uploadOptions.buttonText}}' \
+						processing-text='{{$ctrl.uploadOptions.processingText}}' \
+						complete-text='{{$ctrl.placeholder}}' \
+						cancel-text='{{$ctrl.uploadOptions.cancelText}}' \
+						too-larger-message='{{$ctrl.uploadOptions.tooLargeMessage}}' \
+						accept='{{$ctrl.uploadOptions.accept}}' \
```
StevePole: I don't think accept and max-size are global.  Though unlikely, I can imagine a form in which they were different.

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
```diff
@@ -126,11 +127,16 @@
...
-						instructions='Upload a file' \
-						button-text='Upload' \
-						accept='' \
+						icon='{{$ctrl.uploadOptions.icon}}' \
+						instructions='{{$ctrl.uploadOptions.instructions}}' \
+						button-text='{{$ctrl.uploadOptions.buttonText}}' \
+						processing-text='{{$ctrl.uploadOptions.processingText}}' \
+						complete-text='{{$ctrl.placeholder}}' \
+						cancel-text='{{$ctrl.uploadOptions.cancelText}}' \
+						too-larger-message='{{$ctrl.uploadOptions.tooLargeMessage}}' \
```
StevePole: Typo: 'larger'

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
```diff
@@ -126,11 +127,16 @@
...
-						icon='upload' \
 						description='{{$ctrl.placeholder}}' \
-						instructions='Upload a file' \
-						button-text='Upload' \
-						accept='' \
+						icon='{{$ctrl.uploadOptions.icon}}' \
+						instructions='{{$ctrl.uploadOptions.instructions}}' \
+						button-text='{{$ctrl.uploadOptions.buttonText}}' \
+						processing-text='{{$ctrl.uploadOptions.processingText}}' \
+						complete-text='{{$ctrl.placeholder}}' \
```
StevePole: Seems weird to use placeholder for this, it could be something like "Please upload photo"

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
```diff
@@ -126,11 +127,16 @@
 				<div ng-switch-when='upload'> \
 					<tw-upload \
 						name='{{$ctrl.name}}' \
-						icon='upload' \
 						description='{{$ctrl.placeholder}}' \
-						instructions='Upload a file' \
-						button-text='Upload' \
-						accept='' \
+						icon='{{$ctrl.uploadOptions.icon}}' \
+						instructions='{{$ctrl.uploadOptions.instructions}}' \
```
StevePole: Don't think 'instructions' should be global, would expect it to use "help text" (or something) from the field.  Not sure if we have something for that in the API yet.

partials/tw-dynamic-control.html
```diff
@@ -71,7 +72,8 @@ <h4 id="dynamic">Dynamic form control</h4>
   ng-max="{{vm.dynamic.max}}"</span></span>
   ng-change='...'
   ng-focus='...'
-  ng-blur='...'
+  ng-blur='...'<span ng-if="vm.dynamic.type === 'upload'">
+  custom-parameters="{{vm.dynamic.customParameters}}"</span>
```
StevePole: Incorrect naming.

partials/tw-dynamic-control.html
```diff
@@ -28,6 +28,7 @@ <h4 id="dynamic">Dynamic form control</h4>
 				ng-change="vm.log('dynamic change')"
 				ng-focus="vm.log('dynamic focus')"
 				ng-blur="vm.log('dynamic blur')"
+				custom-parameters="vm.dynamic.customParameters"
```
StevePole: Incorrect naming.

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
StevePole: I would thing max-size should use the 'max' from the field.  Perhaps if there was some 'format' option for fields that would also apply here for the types...?

lapostoj: Not sure what you call "global" here and in the other comment, with that structure, if there are 2 uploads in the same form they could have different values.

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
lapostoj: image
image

So here my reasoning was that if you look at the screenshots, "placeholder" is the "Front of your ID document" on the first and "complete-text" is "Front of ID" on the second.
So I see no reason (at least in how they are used in this example) why they would be different, it basically states the expected content, which I think make sense to have both before and after picking a file.

partials/tw-dynamic-control.html
```diff
@@ -28,6 +28,7 @@ <h4 id="dynamic">Dynamic form control</h4>
 				ng-change="vm.log('dynamic change')"
 				ng-focus="vm.log('dynamic focus')"
 				ng-blur="vm.log('dynamic blur')"
+				upload-optins="vm.dynamic.uploadOptions"
```
StevePole: Typo :-)

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
StevePole: What about this one?

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
StevePole: I would think on the final page it would be more appropriate to use the 'label' of the field, because labels are normally indicative, they state what is required without instruction.  Whereas placeholders are normally imperative, they instruct what to do.

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
StevePole: Actually, instructions seems more like what a placeholder would do?

StevePole:
Longer term think we should change the twUpload interface to be called placeholder and help-text.  Need to migrate clients though.  Might support both initially and update docs to push people towards better option,

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
```diff
@@ -19,17 +19,23 @@
 				name: "@",
 				id: "@",
 				placeholder: "@",
+				helpText: "@",
 				step: "@",
 				locale: "@",
+				accept: "@",
+				icon: "@",
+				tooLargeMessage: "@",
```
lapostoj: Yes had the same thought, will add them!

angular/components/tw-dynamic-form-control/tw-dynamic-form-control.directive.js
```diff
@@ -19,17 +19,23 @@
...
 				options: "=",
+				max: "=",
```
lapostoj: I think here we should more push toward the "max" and deprecate the "ngMax"... It makes no sense for mobile clients to have a parameter called "ngMax"...

StevePole: I don't mean in the API, just the component.  Same as API doesn't have ngMaxLength, it has maxLength, if numeric the component has ngMax, API has max.  Main aim of these was to closely mimic what a native control would have.  So in Angular you do:

```
<input type="text" ng-max-length="10" />
```

Not sure if there is something for <input type="file" ...

lapostoj: Hmm yeah, I guess here it was the idea in mind to keep the naming consistent across the different level, but would also work with ngMax, ok!

lapostoj:
After we are done with this release I'll try to go through and at least mark as deprecated a number of stuff which we should probably change in the interfaces to initiate that 🙂

PR #77, opened 2017-04-26
[DON'T MERGE] Allow copying upload url

keyvanakbary:
In projects like Lienzo or the Blog, we not only need to upload a file, but also to display the file path to be copied by whoever is uploading the file. I've added this possibility to the uploader. If you setup the `copy-upload-url-text="Copy"` property in in the directive, an input will appear to be copied.

image

The component knows the final upload URL because it process the response from the server, basically it just stores and then displays what the server returns as a response when uploading the file.

The copy functionality is a little bit tricky and I am a bit unsure on how to put this as a dependency. I just placed this thing completely outside the component, so if you want to enable the copy functionality, you'll have to do something like this

```js
angular.module('my-app', ['tw.styleguide-components', 'ngclipboard'])
```

and install the `ngclipboard` package by your own. I'm not sure if this is the correct way of doing this. I hope to get some feedback.

This changeset shouldn't affect any client, only if you have that special option, the upload copy input should appear.


keyvanakbary:
I've implemented this logic out of the component to avoid specific use case logic in here 👍

PR #78, opened 2017-04-27
Fix http response not being passed

keyvanakbary:
`response` was never being passed to the `onSuccess` handler because there was no return statement with it. Fixed it 👍


angular/components/tw-upload/tw-upload.directive.js
```diff
@@ -192,6 +192,7 @@
 					])
 					.then(function(response) {
 						showDataImage(response[1]);
+						return response[0];
```
lapostoj: worth add/updating the test for it? 🙈

lapostoj: Also sure we need only `response[0]` and not `response`? Not sure what the different parts contain.

keyvanakbary: 1) It looks like the tests cannot be run because they are breaking
2) `response[0]` is the first queued promise, the one referring to the HTTP request (basically the response), the second one is about reading the file (the byte code). Following what we say in the docs, the HTTP response should be passed to the `onSuccess` handler and that does not happen currently. I do not think it makes much sense to pass the read file to the handler as the primary action is not reading from it for the preview but actually uploading it.

lapostoj: Ok will check what is happening with the tests 🤔 

2 makes sense, mostly was to understand better 🙂

keyvanakbary:
@lapostoj I bumped the version but no new artifacts where generated. I guess this relies exclusively in git tags 👍

lapostoj:
Yes, the artifacts only come from the github release, I just guess it is better if we can keep it in sync 🙂

PR #84, opened 2017-05-19
Add validation/error message options to requirement form

StevePole:
Also add width option for column layout


angular/components/tw-requirements-form/tw-dynamic-form-section.directive.js
```diff
@@ -9,68 +9,78 @@
...
 				model: '=',
-				uploadOptions: '=',
+				fields: '<',
+				uploadOptions: '<',
 				locale: '@',
-				onRefreshRequirements: '&'
+				onRefreshRequirements: '&',
+				validationMessages: '<',
+				fieldErrors: '<',
+				isValid: '=?'
```
lapostoj: this is to send back up an information about the validity of the form, right?

StevePole: Yep, just toggles between true/false depending on validity of the fields inside.  The fields won't necessarily be styled as invalid as some start life without 'red' styling and becomes red after interacted with and blurred (or after form submit).

PR #85, opened 2017-05-25
Change formSection to fieldset, rename fieldErrors to errorMessages

StevePole:


angular/components/tw-requirements-form/tw-fieldset.directive.js
```diff
@@ -3,94 +3,98 @@
...
-										ng-max='field.max' \
-										ng-pattern='field.validationRegexp' \
-										tw-validation \> \
-										<!-- tw-dynamic-async-validator='field.validationAsync' --> \
-									</tw-dynamic-form-control> \
-									<div class='error-messages'> \
-										<div ng-repeat='(validationType, validationMessage) in $ctrl.validationMessages' \
-											class='error-{{validationType}}'> \
-											{{validationMessage}} \
+				<fieldset ng-form='twFieldset'> \
```
lapostoj: So we should aim at using these "fieldset" tags more?

StevePole: Not sure we will need them more than this, but makes sense to use markup with a defined purpose where possible, as it will be beneficial to screen readers.  I also added an optional 'legend' parameter which is another native tag for describing the purpose of the fieldset, in some of our custom forms we already use this kind of grouping, and I think it will be necessary in future: https://docs.google.com/document/d/1bHdHFMOqpIKXJOGXM8rJkJ57e7sGZTKQOO6laUDJ9iA/edit#heading=h.vpb2x3wglwgl

StevePole:
I will bump version when ready to merge, as other PRs inbound.

PR #86, opened 2017-05-30
HUF Decimal Input (CARDS-2373)

mscott-tw:
Set the decimal places allowed for HUF to 0. While technically the currency does support decimals, due to the exchange rate banks do not accept amounts with decimals, so we will do the same and only allow whole numbers.


angular/services/tw-currency-data/tw-currency-data.service.js
```diff
@@ -23,6 +23,7 @@
 			'XAF': 0,
 			'XOF': 0,
 			'XPF': 0,
+			'HUF': 0,
```
StevePole: Could you add a comment for why we're deviating from ISO?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tw-styleguide-components",
-  "version": "1.16.0",
+  "version": "1.17.0",
```
StevePole: I think we could make this a patch release.

PR #87, opened 2017-06-07
Does not select first value in tw-select if placeholder is present

timasjov:
- In case of placeholder present do not select the first value of the list. 
- In case of missing placeholder select first value from the list.


angular/components/tw-select/tw-select.directive.js
```diff
@@ -495,7 +495,7 @@
 
 	function setDefaultIfRequired($ngModel, $ctrl, $element, $attrs) {
 		// If required and model empty, select first option with value
-		if (($ctrl.ngRequired || $attrs.required) && !isValidModel($ctrl.ngModel)) {
+		if (($ctrl.ngRequired || $attrs.required) && !isValidModel($ctrl.ngModel) && !$ctrl.placeholder) {
```
lapostoj: Really much of a detail but I'd probably put it first as it's the easiest condition to test (doesn't make you test the rest if it's false)

PR #89, opened 2017-06-09
Removes fieldset error on onBlur and onChange events

timasjov:
If server returns an error, it is properly rendered next to the `tw-dynamic-form-control` field inside `tw-fieldset` and highlighted in red. However, if you change the input and server returns error again, then it will not be rendered. Current PR fixes this behaviour by removing field error inside the `onChange` and `onBlur` events.


timasjov:
Formatting is a bit off, will fix.

angular/components/tw-requirements-form/tw-fieldset.directive.spec.js
```diff
@@ -0,0 +1,103 @@
...
+                  "key": "sortCode",
+                  "type": "text",
+                  "refreshRequirementsOnChange": false,
+                  "required": true,
+                  "displayFormat": "**-**-**",
+                  "example": "40-30-20",
+                  "minLength": 6,
+                  "maxLength": 8,
+                  "validationRegexp": null,
+                  "validationAsync": {
```
StevePole: Could you remove the validationAsync bit?  It's not relevant to the tests, and it may break in future as it will cause an HTTP request that is not mocked.

PR #93, opened 2017-06-16
Add field formatting

StevePole:


.gitignore
```diff
@@ -5,6 +5,7 @@
 /build/
 /components/
 npm-debug.log
+package-lock.json
```
skywritergr: We normally add this to our global gitignore and not to the project gitignore.

angular/styling/tw-text-format/tw-text-format.service.js
```diff
@@ -0,0 +1,111 @@
...
+        value = "";
+      }
+      if (typeof pattern !== "string") {
+        return value;
+      }
+      var newValue = "";
+      var separators = 0;
+      var moveCursor = 0;
+
+      for(var i = 0; i < pattern.length && i <= value.length + separators; i++) {
```
skywritergr: This conditional in the for loop it's a bit too elaborate for my liking. I like the conditionals in the for loops to be simple. Worth exporting it somehow?

angular/styling/tw-text-format/tw-text-format.service.js
```diff
@@ -0,0 +1,111 @@
...
+      // Add remaining characters on the end without formatting
+      if (value.substring(pattern.length - separators, value.length)) {
+        newValue += value.substring(pattern.length - separators, value.length);
+      }
+      return newValue;
+    };
+
+    this.unformatUsingPattern = function(value, pattern) {
+      if (!value) {
+        return "";  // return value
```
skywritergr: Is this what you intent? What's the deal with the comment?

angular/styling/tw-text-format/tw-text-format.service.js
```diff
@@ -0,0 +1,111 @@
...
+      while (pattern[position - separators - 1] &&
+        pattern[position - separators - 1] !== "*") {
+        separators++;
+      }
+      return separators;
+    };
+
+    this.countSeparatorsAfterCursor = function(pattern, position) {
+      var separators = 0;
+      while (pattern[position + separators] &&
```
skywritergr: Again, like the for loop above. This is slightly more complicated for a loop than i like.

angular/styling/tw-text-format/tw-text-format.service.js
```diff
@@ -0,0 +1,111 @@
...
+    };
+
+    // How long will a value be after separators have been inserted
+    this.countSeparatorsInAppendedValue = function(pattern, position, value) {
+      var separators = 0;
+      var i = 0;
+      var toAllocate = value.length;
+      while(toAllocate) {
+        if (pattern[position + i] === "*" ||
+          typeof pattern[position + i] === "undefined") {
```
skywritergr: is it OK if it's null for example?

partials/tw-text-format.html
```diff
@@ -0,0 +1,76 @@
...
+  <div class="col-md-6">
+    <div class="form-group">
+      <label class="control-label">Text format</label>
+      <input type="text" class="form-control"
+        ng-model="$ctrl.model"
+        ng-minlength="$ctrl.pattern.minlength"
+        ng-maxlength="$ctrl.pattern.maxlength"
+        ng-required="true"
+        placeholder="{{$ctrl.pattern.placeholder}}"
+        tw-focusable
```
skywritergr: Does that mean that the field can be edited? Isn't it better to name this "tw-readonly" or something? Focusable is the default action of any html element

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+  }
+
+  function TwTextFormatController(
+    $element,
+    $timeout,
+    $scope,
+    TwTextFormatService,
+    TwUndoStackFactory
+  ) {
+    var ngModelController, element, undoStack, keydownCount,
```
skywritergr: It's a better idea to do one var per line for future maintenability. Babel and Webpack and ESLint etc are favouring that approach.

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+
+      $scope.$watch('$ctrl.twTextFormat', onPatternChange);
+      $scope.$watch('$ctrl.ngModel', onModelChange);
+
+      // Make sure initial pattern is setup
+      onPatternChange($ctrl.twTextFormat);
+
+      // We need the formatter for external model changes
+      ngModelController.$formatters.push(function(value) {
+        return TwTextFormatService.formatUsingPattern(value, pattern); //$ctrl.twTextFormat);
```
skywritergr: Is the comment necessary?

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+      });
+      ngModelController.$parsers.push(function(value) {
+        return TwTextFormatService.unformatUsingPattern(value, pattern); //$ctrl.twTextFormat);
```
skywritergr: Same as above. Comment.

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+      }
+      if (newPattern) {
+        viewValue = TwTextFormatService.formatUsingPattern(viewValue, pattern);
+      }
+      undoStack.reset(viewValue);
+      element.value = viewValue;
+    }
+
+    function replaceLengthValidators(ngModelController) {
+      $timeout(function () {
```
skywritergr: Why do we need a timeout?

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+        var originalMinLength = ngModelController.$validators.minlength,
+            originalMaxLength = ngModelController.$validators.maxlength;
+
+        if (originalMinLength) {
+          ngModelController.$validators.minlength = function(modelValue, viewValue) {
```
skywritergr: I don't understand what is happening here. We are conditionally creating a function?

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+            return originalMinLength(
+              modelValue,
+              TwTextFormatService.unformatUsingPattern(viewValue, pattern)
+            );
+          };
+        }
+        if (originalMaxLength) {
```
skywritergr: Same question as above. I don't thing this is the right approach here. Provisions for a case of null or undefined should be inside the function (that should always be there). Conditionally declaring a variable is not great. It's like an API point that sometimes is there and sometimes not.

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+            return originalMaxLength(
+              modelValue,
+              TwTextFormatService.unformatUsingPattern(viewValue, pattern)
+            );
+          };
+        }
+      });
+    }
+
+    function reformatControl(element, originalValue) {
```
skywritergr: Seems like this function over here is doing two different things. Returning the new value and setting the element value. If that's the case indeed the name of the function is not accurately depicting this. I would prefer to split it into two functions.

angular/styling/tw-text-format/tw-text-format.directive.js
```diff
@@ -0,0 +1,419 @@
...
+      keys.alt,
+      keys.left,
+      keys.up,
+      keys.right,
+      keys.down
+    ];
+
+    init();
+
+    function getPositionAfterBackspace(pattern, element, selectionStart, selectionEnd) {
```
skywritergr: Can we move these function up to be grouped with the other ones that do similar functions?

skywritergr:
Also one comment regarding tests and the cursor movement. Have you tried the chromedriver? They have a similar situation like phantom and firefox

.gitignore
StevePole: I don't follow, what global ignore?

angular/styling/tw-text-format/tw-text-format.service.js
StevePole: We're accessing a character in a string, so can only be a character or undefined if the string isn't long enough.  I could change to check vs the string length instead, might be clearer?

angular/styling/tw-text-format/tw-text-format.service.js
StevePole: Will remove 👍

partials/tw-text-format.html
StevePole: No, this is a directive that hoists the focus class up to the parent form-group.  Not new in this PR.

angular/styling/tw-text-format/tw-text-format.directive.js
StevePole: 👍

angular/styling/tw-text-format/tw-text-format.directive.js
StevePole: 👍

angular/styling/tw-text-format/tw-text-format.directive.js
StevePole: 👍

angular/styling/tw-text-format/tw-text-format.directive.js
StevePole: Will add comment

angular/styling/tw-text-format/tw-text-format.directive.js
StevePole: We are replacing the minlength validator only if there is a minlength validator.

angular/styling/tw-text-format/tw-text-format.directive.js
StevePole: 👍

angular/styling/tw-text-format/tw-text-format.service.js
skywritergr: Yeah probably. In general the typeof situation is not widely used. But it's not mistaken by any means.

.gitignore
skywritergr: In git you can ignore files in two ways. 
1. a .gitignore file in the project folder that you commit
2. a global ignore on your computer that it's not committed. 

In the global ignore on your machine you should have ignores for yarn.lock or npm-lock.json and maybe other files unique to your environment.

https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

skywritergr:
Also can we squash all the commits?

StevePole:
I normally squash at point of merge.

StevePole:
Chrome and Safari are switched on for cursor tests, firefox and phantom are disabled.

PR #94, opened 2017-06-28
Expandable cards

yashnevatiatw:


index.js
```diff
@@ -331,4 +331,7 @@ angular.module('my-app', ['tw.styleguide-components'])
 })
 .directive('twProcessDocs', function() {
 	return {templateUrl: 'partials/tw-process.html'};
+})
+.directive('twCardsDoc', function(){
```
StevePole: 'Docs'

GruntFile.js
```diff
@@ -122,6 +124,32 @@ module.exports = function(grunt) {
                     mangle: true,
                     beautify: false
                 }
+            },
+            layoutComponents: {
+                src: [
+                    'angular/layout-components/**/*.controller.js',
+                    'angular/layout-components/**/*.directive.js',
+                    'angular/**/*.service.js'
```
StevePole: Don't think you need services in here.

angular/layout-components/tw-cards/tw-cards.directive.js
```diff
@@ -0,0 +1,154 @@
...
+                        <transfer-type-icon type="$ctrl.icon"></transfer-type-icon> \
+                    </div> \
+                </div> \
+                <div class="media-body" ng-transclude="collapsedCard"></div> \
+            </div> \
+        </div>';
+
+    var expandedCardTemplate = 
+        '<div   class="collapse" \
+                ng-attr-aria-expanded = { $ctrl.expanded } \
```
StevePole: Looks like you're missing some quotes on this attribute.

StevePole: Also, can we remove the whitespace `ng-attr-aria-expanded="{{ $ctrl.expanded }}"`?

angular/layout-components/tw-cards/tw-cards.directive.js
```diff
@@ -0,0 +1,154 @@
...
+    function CardContainerController(){
+
+        var $ctrl = this;
+        $ctrl.cards = [];
+        $ctrl.expandedIdx = null; // tracks the expanded card
+        $ctrl.addCard = function(expanded) {
+            $ctrl.cards.push(expanded);
+        };
+        $ctrl.toggle = function(index) {
+            if($ctrl.expandedIdx != null && $ctrl.expandedIdx !== index){
```
StevePole: `if {`

angular/layout-components/tw-cards/tw-cards.directive.js
```diff
@@ -0,0 +1,154 @@
...
+        };
+    }
+
+    function CardContainer() {
+        return {
+            scope: {
+                inactive: '=?'
+            },
+            controllerAs: '$ctrl',
+            controller: [CardContainerController],
```
StevePole: No need for array if we have nothing to inject, just `CardContainerController`

angular/layout-components/tw-cards/tw-cards.directive.spec.js
```diff
@@ -0,0 +1,33 @@
...
+                </tw-cards>';
+        }
+        var element = angular.element(template);
+        // append to document so we can test document.activeElement
+        angular.element(document.body).append(element);
+        var compiledElement = $compile(element)($scope);
+
+        $scope.$digest();
+        return compiledElement;
+    }
```
StevePole: No actual tests :-), can pair on one if that helps?

angular/layout-components/tw-cards/tw-cards.directive.js
yashnevatiatw: what's off here ?

StevePole: Sorry, just a style thing, no space between if and the brace `if(` vs `if (`.

angular/layout-components/tw-cards/tw-cards.directive.spec.js
yashnevatiatw: forgot to finish them, just added a couple

angular/layout-components/tw-cards/tw-cards.directive.spec.js
```diff
@@ -0,0 +1,88 @@
...
+                            <a class='transcluded-content'></a> \
+                        </expandin> \
+                    </tw-card> \
+                </tw-cards>";
+            directiveElement = getCompiledDirectiveElement($scope, template);
+        });
+        it('should render transcluded content of collapsed card', function() {
+            var transcluded = directiveElement.find('.transcluded-content');
+            expect(transcluded.length).toBe(1);
+        });
```
StevePole: Can you check that collapsed section is not rendered?

angular/layout-components/tw-cards/tw-cards.directive.spec.js
```diff
@@ -0,0 +1,88 @@
...
+                        colour='blue' \
+                        expanded='true'> \
+                        <collapsed> \
+                            <a class='transcluded-content'></a> \
+                        </collapsed> \
+                        <expandin> \
+                            <a class='transcluded-content'></a> \
+                        </expandin> \
+                    </tw-card> \
+                </tw-cards>";
```
StevePole: Looks like both tests use the same template, maybe just make this the default template?

StevePole: I see the difference, you could do that with a scoped variable.  Just change the value in the scope, and force a $scope.$apply()

angular/form-components/tw-upload/tw-upload.directive.spec.js
```diff
@@ -21,16 +21,17 @@ describe('Directive: TwUpload', function() {
...
+                <tw-cards> \
+                    <tw-card \
+                        colour='blue' \
+                        buttonText='or click here' \
+                        placeholder='please choose' \
+                        on-upload='onUpload' \
+                        ng-accept='csv'> \
+                        <a class='transcluded-content'></a> \
+                    </tw-card> \
+                </tw-cards>";
```
StevePole: Think this is an error! :-)

angular/layout-components/tw-cards/tw-cards.directive.spec.js
StevePole: Could you add a couple more tests for:
* what happens when two cards are both given the expanded=true?
* is it possible to link two sets of card, such that opening a card closes the open card in the other list?

partials/tw-cards.html
```diff
@@ -0,0 +1,586 @@
...
+                </div>
+                <div class="col-sm-12 col-lg-4 p-t-3 m-b-0">
+                    <button class="btn btn-success btn-block">Repeat transfer</button>
+                </div>
+            </div>
+        </card-form>
+    </tw-card>
+    <tw-card
+        colour = "none"
+        icon = "TRANSFER"
```
StevePole: Just looking at the Borderless page, I think we need a little more flexibility around icons:

image

Could we do this better by transcluding the icon in?

StevePole: If we did that I would want the circle to be part of the template, and the content to be transcluded, not sure what's possible.  The other alternative would be to support a flag property as an alternative, but I think first option might be better.

angular/layout-components/tw-cards/tw-cards.directive.js
```diff
@@ -0,0 +1,154 @@
...
+                icon: '@',
+                form: '<',
+                index: '<',
+                expanded: '=?',
+                disabled: '=?',
+                inactive: '<'
+            },
+            transclude: {
+                collapsedCard: 'collapsed',
+                expandedCard: 'expandin',
```
StevePole: Can we rename this, expandin isn't a word :-)  Would much prefer expanded as it mirrors well with collapsed

angular/services/tw-cards-service/tw-cards-service.service.js
```diff
@@ -0,0 +1,45 @@
...
+                expandedIndex = index;
+                cards[index].enlarged = true;
+            }
+        };
+
+        this.addCard = function(scope) {
+            cards.push(scope);
+        };
+
+        this.getEI = function(){
```
StevePole: Better to avoid contractions on variable names, as it's difficult for others to understand your code.  Could you change to getExpandedIndex

partials/tw-cards.html
```diff
@@ -0,0 +1,597 @@
...
+                        <span class="sr-only">Received</span> £10.00 <span class="hidden-xs ng-binding">GBP</span>
+                    </p>
+                </div>
+            </div>
+        </collapsed>
+        <expanded>
+            <div>
+                <div class="m-t-1 m-b-3 visible-xs-block visible-sm-block">
+                    <p class="h2 m-b-0">
+                        <transfer-status-message class="ng-isolate-scope">
```
StevePole: Lots of these unnecessary tags still around, maybe you deleted the one with the comment?  Would be good to delete all, as they're not doing anything.

angular/layout-components/tw-cards/tw-cards.directive.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    beforeEach(inject(function($injector) {
+        $rootScope = $injector.get('$rootScope');
+        $compile = $injector.get('$compile');
+        $scope = $rootScope.$new();
+    }));
+
+    var cardTemplate = "\
+        <tw-card \
+            colour='blue' \
+            enlarged='enlarged'> \
```
StevePole: In the service you use expanded, here you use enlarged, which is likely to cause confusion.  I'm guessing Angular complains if you call this property expanded as well as the transclusion slot?  Perhaps active would be a better name to use across this and the service?

angular/services/tw-cards-service/tw-cards-service.service.js
```diff
@@ -0,0 +1,45 @@
...
+        	expandedIndex = newEI;
+        };
+
+        this.getCard = function(index){
+        	return cards[index];
+        };
+
+        this.getLength = function(){
+            return cards.length;
+        };
```
StevePole: I'm wondering if we can do a bit better using the service.  I feel we're mixing responsibilities a bit.  What if each card took an id property, and then registered that id with the card service (it could use the index, but I'm a bit worried how that works across different card sets)?  The service could provide an isActive(id) method and toggle(id) to manage open/closing.  I think then you could remove your link function on the card directive, and simplify things a fair bit?

partials/tw-cards.html
```diff
@@ -0,0 +1,597 @@
+<h4 id="cards">Cards</h4>
+<p>These are cards that have information about an activity and provide
+    more detail when they're expanded.</p>
+<p>They expand differently on mobile</p>
+<hr />
+
+<tw-cards>
+    <tw-card
+        form = "false"
+        colour = "red"
```
StevePole: I'm thinking about whether we should use the Bootstrap states rather than colours, e.g. state: warning, danger, info, success...

angular/layout-components/tw-cards/tw-cards.directive.spec.js
yashnevatiatw: yeah, can't use expanded for both; 
using 'active' / 'inactive' to differentiate between active and past activities (i.e. completed transactions

partials/tw-cards.html
```diff
@@ -0,0 +1,490 @@
...
+</tw-cards>
+<tw-cards inactive="true">
+    <tw-card
+        form="true"
+        state="none"
+    >
+        <card-icon>V</card-icon>
+        <collapsed>
+            <div class="media">
+                <div class="media-body">
```
StevePole: Could we build a couple small components/directives for the content of the cards?  We have a lot of duplication,  Would be cool if this markup looked more like:
```
<tw-cards>
  <tw-card>
    <collapsed><transfer-heading></transfer-heading></collapsed>
    <expanded><transfer-expanded></transfer-expanded></collapsed>
  </tw-card>
  <tw-card>
    <collapsed><example-transfer-heading></example-transfer-heading></collapsed>
    <expanded><example-transfer-expanded></example-transfer-expanded></collapsed>
  </tw-card>
  ...
</tw-cards>
```
This is more representative of how it will actually look once integrated.

partials/tw-cards.html
```diff
@@ -0,0 +1,490 @@
+<h4 id="cards">Cards</h4>
+<p>These are cards that have information about an activity and provide
+    more detail when they're expanded.</p>
+<p>They expand differently on mobile</p>
+<hr />
+
+<tw-cards>
```
StevePole: Can we put the component inside a http://bootstrap.transferwise.com/components/#wells so we get the grey background?

partials/tw-cards.html
```diff
@@ -0,0 +1,490 @@
...
+                <div class="m-t-2 btn-toolbar " transfer="$ctrl.transfer">
+                    <button class="btn btn-primary ">Pay for this transfer</button>
+                    <button class="btn btn-primary ">I've already paid</button>
+                    <button class="btn btn-danger pull-md-right m-b-0">Cancel transfer</button>
+                </div>
+            </div>
+        </expanded>
+    </tw-card>
+</tw-cards>
+<tw-cards inactive="true">
```
StevePole: Could you add some spacing between the two section, maybe put a title "Inactive" between them?

partials/tw-cards.html
```diff
@@ -0,0 +1,490 @@
...
+                    <button class="btn btn-success btn-block">Repeat transfer</button>
+                </div>
+            </div>
+        </card-form>
+    </tw-card>
+    <tw-card
+        state="none"
+        form="true"
+    >
+        <card-icon>W</card-icon>
```
StevePole: Could we have a couple examples with an actual icon and maybe an image as well (to represent the flag scenario)?

angular/layout-components/tw-cards/tw-cards.directive.js
```diff
@@ -0,0 +1,203 @@
...
+                <div> \
+                    <div class="m-t-1 m-b-3 visible-xs-block visible-sm-block" ng-transclude="expHead"></div> \
+                    <div ng-transclude="expBody"></div> \
+                    <hr class="hidden-xs hidden-sm hidden-md"> \
+                    <div class="m-t-2 btn-toolbar" ng-transclude="expButt"></div> \
+                </div>'
+        };
+    }
+
+    function FrmExample(){
```
StevePole: What are these examples?  Are they for the demo page?  They shouldn't be in the directive file as this will be packaged for live, I would move to index.js.

StevePole: Also, don't contract names: 'CollExample', 'ExpExample', not as clear as 'CollapsedExample', 'ExpandedExample'.

angular/layout-components/tw-cards/tw-cards.directive.spec.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
+
+fdescribe('Directive: TwCards', function() {
```
StevePole: Remove 'f'?

PR #95, opened 2017-07-13
Adds form attribute to tw-card and fixes the demo

yashnevatiatw:


StevePole:
What was this for?  Do we still need the changes, master has changed quite dramatically now.

PR #101, opened 2017-08-07
Add options to hide the label in the selected element and option to drop upwards

Tankenstein:
Needed for transfer categories component.


angular/form-components/tw-select/tw-select.directive.js
```diff
@@ -25,15 +25,17 @@
 				size: '@',
 				dropdownRight: '@',
 				dropdownWidth: '@',
+				dropdownUp: '@',
```
OFFLlNE: I would've put that on line 27, but that's fine 😂

Tankenstein: Good call :D

angular/form-components/tw-select/tw-select.directive.js
```diff
@@ -25,15 +25,17 @@
...
 				inverse: '=',
 				hideNote: '@',
 				hideSecondary: '@',
 				hideIcon: '@',
 				hideCurrency: '@',
-				hideCircle: '@'
+				hideCircle: '@',
+				hideLabel: '@'
 			},
 			template: " \
```
OFFLlNE: wouldn't it be better to create a template html for that? Easier to handle in the future if needed too

partials/tw-select.html
```diff
@@ -103,13 +109,15 @@ <h4 id="select">TW Select</h4>
 				filter="{{$ctrl.select.showFilter}}"
 				size="{{$ctrl.select.size}}"
 				dropdown-right="{{$ctrl.select.dropdownRight}}"
+				dropdown-up={{$ctrl.select.dropdownUp}}
```
OFFLlNE: this does not need doublequotes around that? Seems like every other-one has

Tankenstein: oopsie

angular/form-components/tw-select/tw-select.directive.js
Tankenstein: Template for what?

oliverviljamaa: I believe he meant a separate `html` file for the template, rather than inlining it. Probably every template in this repo is done like this as adding the `html` contents to template caches is not implemented yet.

StevePole: Went this direction early on to avoid relative pathing issues.  Happy to hear other ideas.

Tankenstein: I like keeping the template next to code (it's one unit), so i think this is fine

PR #104, opened 2017-08-21
Add an option to show a shorter date

SofiiaHorishna:


angular/form-components/tw-date-lookup/tw-date-lookup.directive.js
```diff
@@ -402,6 +403,10 @@
 				}
 			});
 
+			$scope.$watch('$ctrl.shortDate', function(newValue, oldValue) {
+				ngModelCtrl.$validate();
```
StevePole: I don't think the short date should affect the model value, only the presentation, so it shouldn't require us to revalidate.

angular/form-components/tw-date-lookup/tw-date-lookup.directive.js
```diff
@@ -24,7 +24,8 @@
 				placeholder: '@',
 				size: '@',
 				locale: '@',
-				label: '@'
+				label: '@',
+				shortDate: '='
```
StevePole: Let's use the '<' one-way binding, as we'll never want to change the value and sync those back out.  It's more efficient.  We don't always use it in the library at the moment as it wa only added to Angular after the library started, but I'm going to move everything (except 2 way stuff like ngModel) to it soon.

angular/form-components/tw-date-lookup/tw-date-lookup.directive.js
StevePole: It would be cool to keep the watch, and to call another function here to update the presentation, there's probably already something you could call but I haven't looked into the code.  This is primarily for the docs, where we might toggle between the presentation formats and we want it to update immediately, but it could also catch bugs on initialisation.  If, for example, the control initialises before a variable that controls the formatting.

angular/form-components/tw-date-lookup/tw-date-lookup.directive.js
```diff
@@ -404,7 +404,7 @@
 			});
 
 			$scope.$watch('$ctrl.shortDate', function(newValue, oldValue) {
-				ngModelCtrl.$validate();
+				updateSelectedDatePresentation();
```
SofiiaHorishna: shortdatewatch

angular/form-components/tw-date-lookup/tw-date-lookup.directive.js
```diff
@@ -493,6 +500,18 @@
                                                                                 $ctrl.locale);
 		}
 
+		function getShortMonthsOfYear(monthsOfYear){
```
StevePole: Could we instead do this in the service instead?  We already pass in 'short', we can just do an additional check if 'short' is the format, but the names are not actually short.

PR #111, opened 2017-08-30
Reformatting and reorganising files for webpack

StevePole:
No functional changes.


angular/form-components.js
```diff
@@ -0,0 +1,44 @@
...
+import Fieldset             from './forms/fieldset/';
+import Focusable            from './forms/focusable/';
+
+import Tabs                 from './navigation/tabs/';
+import Loader               from './loading/loader/';
+import Process              from './loading/process/';
+
+import RequirementsForm     from './forms/requirements-form/';
+
+// Deprecated
```
Tankenstein: If it's deprecated, let's add to the comment what it's replaced with

angular/forms/amount-currency-select/tw-amount-currency-select.component.spec.js
```diff
@@ -13,8 +13,10 @@ describe('Directive: TwAmountCurrencySelect', function() {
   var INPUT_SELECTOR = 'input';
   var SELECT_SELECTOR = 'select';
 
-  beforeEach(module('tw.form-components'));
-  beforeEach(module('tw.form-styling'));
+  //beforeEach(module('tw.styleguide.forms.amount-currency-select'));
```
Tankenstein: Why these comments?

angular/forms/amount-currency-select/tw-amount-currency-select.controller.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    return viewValue <= $ctrl.ngMax;
+  };
+
+  $ctrl.changedAmount = function() {
+    if ($ctrl.ngChange) {
+      // $timeout is needed to get the last ngModel value.
+      // See: https://github.com/angular/angular.js/issues/4558
+      $timeout($ctrl.ngChange);
```
Tankenstein: If you're essentially deferring ngChange to run it once the stack clears, could we maybe define a defer function that would just do $timeout?

angular/forms/date-lookup/tw-date-lookup.component.spec.js
```diff
@@ -7,7 +7,8 @@ describe('Directive: TwDateLookup, ', function() {
       element,
       TwDateService;
 
-  beforeEach(module('tw.form-components'));
+  beforeEach(module('tw.styleguide.forms.date-lookup'));
+  beforeEach(module('tw.styleguide.services.date'));
```
Tankenstein: Should we not rely on the dependencies of the main module you're testing to initialize the dependencies of it?

StevePole: The way I was thinking about this is that it should be as modularised as possible, so you can pick and choose what you want, and so we have flexibility to split the repo at a later date if desired.  Therefore the individual components should specify their dependencies, and the tests should be able to rely on that, rather than a 'global' module.  However, I have deliberately limited changes in the tests so far, and haven't yet looked into using webpack to run the tests, so I'm not 100% sure of this approach.

angular/forms/amount-currency-select/tw-amount-currency-select.controller.js
StevePole: I'm not going to look into that now, haven't made any changes to content of this controller in this PR.

angular/forms/amount-currency-select/tw-amount-currency-select.component.spec.js
StevePole: Didn't seem to work with the more specific modules, hadn't had chance to resolve.  I'll either get it working or remove 👍

angular/forms/amount-currency-select/tw-amount-currency-select.controller.js
Tankenstein: Cool

angular/forms/date-lookup/tw-date-lookup.component.spec.js
Tankenstein: This is exactly what i'm saying, if you're testing 'tw.styleguide.forms.date-lookup' it should specificy the date services as a dependency already if it needs it

StevePole: Ah, yes, agreed.  In practice however, it's not working and I'm not sure why.  In this case the controller of the component imports the service as a dependency, but when you run the tests without this line, it cannot find the date service.  Any ideas what could be up?

PR #112, opened 2017-08-30
Add html-loader, move HTML templates to separate files.

StevePole:


angular/forms/amount-currency-select/tw-amount-currency-select.component.js
```diff
@@ -32,47 +32,7 @@ const TwAmountCurrencySelect = {
...
-          options="$ctrl.currencies" \
-          filter="{{ $ctrl.currencyFilterPlaceholder }}" \
-          ng-change="$ctrl.changedCurrency()"> \
-            <a href="" ng-if="!!$ctrl.customActionLabel" ng-click="$ctrl.onCustomAction()"> \
-              {{ $ctrl.customActionLabel }} \
-            </a> \
-        </tw-select> \
-      </span> \
-    </div>',
+  template: require('./amount-currency-select.html')
```
oliverviljamaa: Why not `import template from './amount-currency-select.html';` and then use the object property shorthand?

StevePole: No particular reason.  I was following a couple examples that did it this way, but your approach is probably better, will change.

oliverviljamaa: We use the import and shorthand in most of our other Angular repos, it's pretty neat:
https://github.com/transferwise/calculator/blob/master/src/calculator/calculator.component.js

oliverviljamaa: bitmoji

PR #113, opened 2017-08-30
Remove tw prefixes

StevePole:


angular/layout/cards/cards.component.js
```diff
@@ -1,12 +1,12 @@
 import template from './cards.html';
 
 const TwCards = {
+  controller: function() {},
```
oliverviljamaa: No need for a NOOP controller.
image

angular/loading/process/process.component.js
```diff
@@ -0,0 +1,14 @@
...
+
+const Process = {
+  bindings: {
+    state: '<',
+    size: '@',
+    onStop: '&',
+    promise: '<'
+  },
+  controller: ProcessController,
+  template: require('./process.html')
```
oliverviljamaa: This seems to require the HTML still.
image

StevePole: Merge fuckup, one sec.

PR #114, opened 2017-08-30
Switch controllers to ES6 class syntax

StevePole:
All controllers switched.


angular/formatting/text-format/text-format.directive.spec.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-describe('Directive: TwTextFormat, ', function() {
+describe('TextFormat directive, ', function() {
```
oliverviljamaa: Do you keep the ` directive` part for actual directives and omit it for components?

karma-conf.js
```diff
@@ -31,7 +31,12 @@ module.exports = function(config) {
         frameworks: ['jasmine'],
 
         // browsers to test against, be sure to install the correct karma browser launcher plugin
-        browsers : ['PhantomJS', 'Chrome', 'Firefox', 'Safari'], // npm install karma-chrome-launcher --save-dev
+        browsers : [
+          'PhantomJS',
```
oliverviljamaa: Hmm, is it even worth testing in `PhantomJS` at all?

angular/forms/amount-currency-select/amount-currency-select.controller.js
```diff
@@ -1,83 +1,85 @@
...
-    return !isNaN(parseFloat(value));
   }
 }
 
-TwAmountCurrencySelectController.$inject = [
+function isNumber(value) {
+  return !isNaN(parseFloat(value));
+}
+
+AmountCurrencySelectController.$inject = [
```
coolsebz: what is your opinion on something like `ngAnnotate` for avoiding to repeat the `$inject` array for each class?
https://github.com/olov/ng-annotate
It would just be an extra step in one of webpack's loaders and gets rid of this repetition 🤔 
We already use it in https://github.com/transferwise/batch-payments-frontend/

angular/formatting/text-format/text-format.directive.spec.js
StevePole: This was a slightly special case.  For text format I provide a directive for formatting inputs, and a filter for string formatting.  They both have tests, so I was just distinguishing them.

karma-conf.js
StevePole: Good question, perhaps not.  The tests are the next thing for an update.  I'm not yet using webpack for the tests, so they don't have module imports yet.  I also want to look at the best setup for circleCI.

angular/forms/amount-currency-select/amount-currency-select.controller.js
StevePole: Interesting, I've not tried it before.  My only reservation is that with the $inject syntax you have the option to map the names.  So you could do:
```
class MyController {
  constructor(InternalNameForThing) { ... }
}
MyController.$inject = ['ExternalNameForThing']
```
Which I've used to move away from the 'public' module name (which prevents clashes with other libraries), and a simpler internal name.

coolsebz: Yep, exactly, it maps dependency names to the same name in the constructor so if that's the common pattern right now then I think it's best to stay consistent and not introduce an additional way of doing things

oliverviljamaa: @coolsebz I've stopped using `ngAnnotate` due to not allowing a discrepancy between the internal and external names and inability to support prefixed services (maybe it's now possible, though).

PR #117, opened 2017-09-05
Add eslint, fix problems, remove jquery usages

StevePole:
Jquery is still required to support the areas where we rely on bootstrap.js, that is:
* dropdowns (inc. keyboard support)
* tooltips
* popovers


src/formatting/text-format/text-format.directive.spec.js
```diff
@@ -20,7 +20,7 @@ describe('TextFormat directive, ', function() {
   describe('Given there is a pattern', function() {
     describe('when typing first character', function() {
       describe('and pattern does not begin with a separator', function() {
-        beforeEach(function () {
+        beforeEach(function() {
           $scope.pattern = "**-**-**";
```
skywritergr: why not single quotes here?

src/formatting/text-format/text-format.directive.spec.js
```diff
@@ -30,21 +30,21 @@ describe('TextFormat directive, ', function() {
           $timeout.flush(1000);
           $scope.$apply();
         });
-        it('should show the character', function () {
+        it('should show the character', function() {
```
skywritergr: I would use an arrow function here.

src/formatting/text-format/text-format.directive.spec.js
```diff
@@ -30,21 +30,21 @@ describe('TextFormat directive, ', function() {
...
           expect(input.value).toBe("1");
         });
-        it('should bind the character to the model', function () {
+        it('should bind the character to the model', function() {
           expect($scope.ngModel).toBe("1");
```
skywritergr: Also, single quotes while we are at it. :)

src/formatting/text-format/text-format.directive.spec.js
```diff
@@ -30,21 +30,21 @@ describe('TextFormat directive, ', function() {
...
         });
-        it('should position the cursor after the character', function () {
+        it('should position the cursor after the character', function() {
           if (supportsCursorPosition()) {
             expect(input.selectionStart).toBe(1);
           }
         });
       });
 
       describe('and pattern begins with a single separator', function() {
```
skywritergr: again, in all those tests I'd use arrow function. If you enable prettier it will do it automatically for you. 👯‍♂️

src/formatting/text-format/text-format.directive.spec.js
```diff
@@ -94,7 +94,7 @@ describe('TextFormat directive, ', function() {
 
     if (supportsCursorPosition()) {
       describe('when cursor', function() {
-        beforeEach(function () {
+        beforeEach(function() {
           $scope.pattern = "***/**/-/**";
```
skywritergr: single quotes everywhere?

src/forms/amount-currency-select/amount-currency-select.controller.js
```diff
@@ -72,7 +78,7 @@ class AmountCurrencySelectController {
 }
 
 function isNumber(value) {
-  return !isNaN(parseFloat(value));
+  return !isNaN(parseFloat(value)); // eslint-disable-line no-restricted-globals
```
skywritergr: Instead of doing that here (or anywhere else you might need it) I'd add an allowed list as displayed here:

https://eslint.org/docs/rules/no-restricted-globals

src/forms/amount-currency-select/amount-currency-select.spec.js
```diff
@@ -60,27 +61,28 @@ describe('AmountCurrencySelect', function() {
...
 
     it('should set ngModel.$dirty when button clicked', function() {
-      input.val('100').trigger('input');
+      $input.val('100').trigger('input');
       expect($ngModel.$dirty).toBe(true);
     });
 
     it('should set ngModel.$touched when blured', function() {
-      input.triggerHandler('blur');
+      input.dispatchEvent(new Event('blur'));
```
skywritergr: Why do this don't have the dollar sign in front of them?

src/forms/amount-currency-select/amount-currency-select.spec.js
```diff
@@ -60,27 +61,28 @@ describe('AmountCurrencySelect', function() {
...
       expect($ngModel.$touched).toBe(true);
     });
 
     it('should style nearest parent form-group when focussed', function() {
-      input.triggerHandler('focus');
-      expect(input.closest('.form-group').hasClass('focus')).toBe(true);
+      input.dispatchEvent(new Event('focus'));
```
skywritergr: the naming here is a bit confusing between `input` and `$input`

src/forms/checkbox/checkbox.controller.js
```diff
@@ -24,54 +39,64 @@ class CheckboxController {
...
       this.$element,
       this.$ngModel,
-      this.ngRequired
+      this.ngRequired,
+      this.dom
     );
   }
 
   // IE 'clicks' the hidden input when label is clicked
+  // eslint-disable-next-line class-methods-use-this
```
skywritergr: I wouldn't do that. If the function isn't using this I'd either remove it outside of the class or mark it as `static`

example: https://eslint.org/docs/rules/class-methods-use-this

src/forms/currency-input/currency-input.spec.js
```diff
@@ -56,26 +57,27 @@ describe('CurrencyInput', function() {
...
 
     it('should set ngModel.$dirty when button clicked', function() {
-      input.val('100').trigger('input');
+      $input.val('100').trigger('input');
       expect($ngModel.$dirty).toBe(true);
     });
 
     it('should set ngModel.$touched when blured', function() {
-      input.triggerHandler('blur');
+      input.dispatchEvent(new Event('blur'));
```
skywritergr: Same thing here, the mix between `input` and `$input` might be confusing. Can we find a different naming convention?

src/forms/date-lookup/date-lookup.controller.js
```diff
@@ -1,11 +1,13 @@
-import TwDateService from '../../services/date/';
+import DateService from '../../services/date/'; // eslint-disable-line no-unused-vars
```
skywritergr: If it is unused why are we importing it?

src/forms/date-lookup/date-lookup.controller.js
```diff
@@ -425,11 +432,19 @@ class TwDateLookupController {
   }
 }
 
-TwDateLookupController.$inject = [
+function resetFocus(element) {
```
skywritergr: why not use an arrow function?

src/forms/date/date.controller.js
```diff
@@ -1,4 +1,4 @@
-import TwDateService from '../../services/date/';
+import DateService from '../../services/date/'; // eslint-disable-line no-unused-vars
```
skywritergr: again, if it's unused why are we importing it?

src/forms/date/date.controller.js
```diff
@@ -257,15 +222,17 @@ function validDate(date) {
 
 function validDateObject(dateObj) {
   return Object.prototype.toString.call(dateObj) === '[object Date]'
-    && !isNaN(dateObj.getTime());
+    && !isNaN(dateObj.getTime()); // eslint-disable-line no-restricted-globals
```
skywritergr: If you set it up as a global exception (or allowance) as mentioned above you can delete these comments.

src/forms/date/date.controller.js
```diff
@@ -257,15 +222,17 @@ function validDate(date) {
...
 }
 
 function validDateString(dateString) {
   return typeof dateString === 'string' && validDateObject(new Date(dateString));
 }
 
 function prepDateLimitForComparison(ngLimit, attrLimit) {
-  var limit = ngLimit ? ngLimit : (attrLimit ? attrLimit : false);
+  // eslint-disable-next-line no-unneeded-ternary
+  let limit = ngLimit ? ngLimit : attrLimit;
```
skywritergr: wouldn't `let limit = ngLimit || attrLimit;` work better? Plus you won't need to ignore the esLint.

src/forms/date/date.controller.js
```diff
@@ -281,14 +248,38 @@ function prepDateValueForComparison(dateValue) {
 }
 
 function extendMonthsWithIds(monthNames) {
-  return monthNames.map(function(monthName, index) {
+  // eslint-disable-next-line arrow-body-style
+  return monthNames.map((monthName, index) => {
```
skywritergr: `(var1, var2) => {value: index, label: monthName}` should work here. No?

src/forms/date/date.controller.js
```diff
@@ -281,14 +248,38 @@ function prepDateValueForComparison(dateValue) {
...
     return {
       value: index,
       label: monthName
     };
   });
 }
 
+function addBlurHandlers($element, $ngModel) {
```
skywritergr: arrow function?

src/formatting/text-format/text-format.directive.spec.js
StevePole: Haven't applied eslint to tests yet

src/formatting/text-format/text-format.directive.spec.js
StevePole: Also haven't moved tests to es6.  Currently the tests are still from grunt, they need to be moved to webpack before I can move to ES6 as that's where my babel step runs.

src/forms/amount-currency-select/amount-currency-select.spec.js
StevePole: What?

src/forms/amount-currency-select/amount-currency-select.controller.js
StevePole: Isn't this a list of things that I want to throw errors?  Currently it's set such that _any_ use of global throws an error.  I've opted in a few case like this where it's a built in function.

src/forms/amount-currency-select/amount-currency-select.spec.js
StevePole: I'm using `$element` for anything that is a jquery (or angular wrapped) HTML node, and just `element` for a pointer to the naked HTML node.  This might go away if jquery is removed from the tests, but that's another huge job.

src/forms/date-lookup/date-lookup.controller.js
StevePole: It is a dependency, it is injected into the controller, so we need that code to be loaded, but Angular handles the DI.  In other cases we take name of the exported module and use that in the $inject, but that's the module name not the service name, which is the thing that needs to be injected, so it doesn't work in this case.

src/forms/date-lookup/date-lookup.controller.js
StevePole: It's a top level function, not a member of a class, so no scope to inherit.  It would still need to be named so not sure of benefit?

src/forms/date/date.controller.js
StevePole: See response above.  Code is used, it's a DI issue.

src/forms/date/date.controller.js
StevePole: Good point 👍

src/forms/date/date.controller.js
StevePole: Another top level function.

src/forms/date-lookup/date-lookup.controller.js
skywritergr: consistency? You are doing it elsewhere in your files anyway.

`const resetFocus = (element) => {...};` would be more consistent

src/formatting/text-format/text-format.directive.spec.js
StevePole: See answer about es6, not able to use it until I move test pipeline to webpack.

src/forms/amount-currency-select/amount-currency-select.spec.js
StevePole: Ah, makes sense in context of following comment, this is an HTML element, $input is the an element wrapped with jquery.

src/forms/checkbox/checkbox.controller.js
StevePole: Yeah, I'll try static 👍

src/forms/date/date.controller.js
StevePole: I'll try 👍

src/services/dom/dom.service.js
```diff
@@ -0,0 +1,60 @@
...
+    return null;
+  };
+}
+
+export default DomService;
+
+
+/*
+
+class DomService {
```
StevePole: Note to self: delete this.

src/services/dom/dom.service.js
```diff
@@ -27,34 +27,3 @@ function DomService() {
 }
 
 export default DomService;
-
-
-/*
```
skywritergr: did we change back to a function or it was never a class?

StevePole: It was never a class, I was trying a couple things, but not currently using classes for services.

src/forms/date/date.controller.js
StevePole: Tried, but didn't work.

PR #118, opened 2017-09-13
Fix bug for tw-tabs onChange callback

vahurh:
With one-way binding you need to specify the callback fn arguments as an object.


src/navigation/tabs/tabs.controller.js
```diff
@@ -8,7 +8,7 @@ class TabsController {
   switchTab(tab) {
     this.active = tab;
     if (this.onChange) {
-      this.onChange(tab);
+      this.onChange( {tab: tab} );
```
lapostoj: probably can use the es6 shorthand and have just `{ tab }`

StevePole:
Could you explain the change in more detail, what does 'properly' mean?  It looks like this changes the object passed, rather than fixing a bug?

Second thing, I have a large number of changes in master that are yet to get a release tag.  I'd like to issue those first under a major release number.  I can try and do that today.

vahurh:
@StevePole because the function was bound with expression binding ('&') instead of '=', you bind it in the html as such `on-change="$ctrl.myOnChange(tab)"` --> when you do that, Angular doesn't know what the argument 'tab' is in the html. Therefore you call the function, explicitly saying what 'tab' is.

Currently it throws an error, see: https://stackoverflow.com/questions/30244358/angularjs-directive-element-method-binding-typeerror-cannot-use-in-operator/30245318

StevePole:
Cool, could you bump the package number, then feel free to merge.

StevePole:
Just rebuild the project and commit to resolve the conflicts as they're just the built files.

PR #121, opened 2017-09-19
Builds dist files for last release, v2.0.5

yashnevatiatw:


src/layout/cards/card.html
```diff
@@ -41,7 +41,7 @@
       <div class="media">
         <div class="media-left">
           <div class="circle circle-sm circle-responsive invisible"></div>
-        </div>
+          </div>
```
StevePole: Whitespace!

PR #128, opened 2017-09-22
Integrate circle ci for tests

StevePole:


circle.yml
```diff
@@ -0,0 +1,32 @@
...
+  override:
+    - yarn
+  cache_directories:
+    - ~/.cache/yarn
+
+test:
+  override:
+    - yarn install; yarn start; yarn test;
+
+# deployment:
```
skywritergr: Delete this.

circle.yml
```diff
@@ -0,0 +1,32 @@
+machine:
+  node:
+    version: 7.10.0
+  environment:
+    YARN_VERSION: 0.24.6
```
oliverviljamaa: I think we can simplify the Yarn part a lot:
https://circleci.com/docs/1.0/yarn/

Also, do we really want Yarn? Npm 5 is pretty much the same thing and comes with Node...

circle.yml
```diff
@@ -0,0 +1,32 @@
...
+        curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version $YARN_VERSION
+      fi
+  override:
+    - yarn
+  cache_directories:
+    - ~/.cache/yarn
+
+test:
+  override:
+    - yarn install; yarn start; yarn test;
```
oliverviljamaa: Any reason why we need the `install` (done in previous step anyway) and `start` (linting could be in `test` and the `demo` commands may not be necessary here)?

package.json
```diff
@@ -42,11 +43,11 @@
   },
   "homepage": "https://github.com/transferwise/styleguide-components",
   "dependencies": {
-    "angular": "~1.5.8",
+    "angular": "<1.6",
```
oliverviljamaa: That doesn't seem accurate, surely the components wouldn't work with `1.0`, for example?

circle.yml
skywritergr: I think that was there more for consistency with other repos, like navigation etc. We can remove it for sure.

StevePole: I copied it from either navigation or frontend-common :man_shrugging:.  No strong preference...

circle.yml
StevePole: George made me do it!  We were hacking around trying to get it to stop failing, turns out the issue was the angular-mocks thing below.  I'll remove.

What do you mean by the demo commands?

package.json
StevePole: Yeah, true. I changed it to this as with the previous setting circle was loading 1.5.x angular and 1.6.x angular-mocks which made the tests fail.  Was just trying to force it.

circle.yml
StevePole: It's literally the next thing that needs to be done, don't see the harm in leaving it there briefly...

circle.yml
oliverviljamaa: Let's keep Yarn then but simplify the setup, WDYT?

karma-conf.dev.js
```diff
@@ -0,0 +1,22 @@
+let karmaConf = require('./karma.json');
+
+karmaConf.plugins = [
```
oliverviljamaa: Double quotes
image

package.json
```diff
@@ -5,11 +5,12 @@
   "license": "MIT",
   "scripts": {
     "pretest": "",
-    "test": "karma start karma-conf.js --single-run",
+    "test": "npm run lint; karma start karma-conf.js --single-run",
+    "test-dev": "karma start karma-conf.dev.js --single-run",
```
oliverviljamaa: Can you explain why we need `test-dev`?

StevePole: That conf tests in Phantom, Chrome, Safari and Firefox.  Wanted to keep the circle config simple to start with, so just went with phantom for that.

circle.yml
StevePole: I'm happy to ditch it, we should just agree what we're using.  I think NPM may be better purely on the basis that people outside the guild may have heard of it...

PR #129, opened 2017-09-25
Remove popover dependency on Bootstrap

StevePole:


src/help/pop-over/pop-over.directive.js
```diff
@@ -1,42 +1,60 @@
...
-  if (tag.getAttribute('data-content-html')) {
-    options.html = true;
+  if (element.dataset.originalTitle) {
+    OPTIONS.title = element.dataset.originalTitle;
+  }
+  if (element.dataset.content) {
+    OPTIONS.content = element.dataset.content;
+  }
+  if (element.dataset.contentHtml) {
+    OPTIONS.html = true; // TODO add support for this
```
StevePole: Did you add support?  I think there may be production usages that require it.

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,337 @@
...
+        elementOffsetDimensions.offsetWidth + POPOVER_SPACING;
+      const popoverOffsetY =
+        (elementOffset.offsetY -
+        (popoverArrowTopOffset + popoverArrowMarginTop + (popoverArrowHeight / 2))) +
+        ((elementOffsetDimensions.offsetHeight / 2) - popoverVerticalBorder);
+
+      popoverOffsets = {
+        offsetX: popoverOffsetX,
+        // offsetY: (offsetY - (popoverOffsetDimensions.offsetHeight / 2))
+        // + (offsetHeight / 2),
```
StevePole: Remove comments if not needed.

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,337 @@
...
+        (elementOffset.offsetY -
+        (popoverArrowTopOffset + popoverArrowMarginTop + (popoverArrowHeight / 2))) +
+        ((elementOffsetDimensions.offsetHeight / 2) - popoverVerticalBorder);
+
+      popoverOffsets = {
+        offsetX: popoverOffsetX,
+        //
+        /* offsetY: (offsetY - (popoverOffsetDimensions.offsetHeight / 2))
+        + (offsetHeight / 2), //Center to center
+        */
```
StevePole: Remove comments if not needed.

src/help/pop-over/pop-over.service.js
mstaicu: Yes, I left those in case we decide to rollback to the initial behaviour of having the popover vertically centered.

src/help/pop-over/pop-over.directive.js
mstaicu: No, not yet. Need more context on this.

StevePole:
Any feedback on tests?  Code looks good but would like to see some before 👍

demo/partials/tw-pop-over.html
```diff
@@ -23,7 +23,7 @@ <h4 id="pop-over">TW Pop Over</h4>
     -
     <a tw-pop-over
       data-original-title="Popover title"
-      data-content-html="true"
+      data-content-html=""
```
skywritergr: What's with the empty string here?

src/help/pop-over/pop-over.directive.js
```diff
@@ -1,42 +1,59 @@
...
+  ELEMENT.setAttribute('data-toggle', 'popover');
+
+  ELEMENT.addEventListener('click', ELEMENT_CALLBACK);
 
-function PopOverLink(scope, element) {
-  if (!element.popover) {
-    // eslint-disable-next-line no-console
-    console.log('twPopOver requires tooltip from bootstrap.js');
-    return;
+  // ELEMENT.addEventListener('mouseover', ELEMENT_CALLBACK);
```
skywritergr: Do we need the comments?

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,366 @@
...
+  let popover = null;
+  let elementWithPopover = null;
+  let popoverPosition = 'right';
+
+  function showPopover(highlightedElement, popoverOptions) {
+    if (highlightedElement instanceof HTMLElement && isOptionsObject(popoverOptions)) {
+      elementWithPopover = highlightedElement;
+      popoverPosition = getPopoverPlacement(popoverOptions);
+
+      const popoverAppendedToBody = Array.from(BODY.children).includes(popover);
```
skywritergr: Array.from isn't supported at all in IE. I hope babel transpiles this.

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,366 @@
...
+      <button class='close popover-close'>&times;</button>
+      <h3 class='popover-title'>__title__</h3>
+      <div class='popover-content'>
+        __content__
+      </div>`;
+  }
+
+  function bindDataToTemplate(template, data) {
+    return template.replace(
+      /__(\w*)__/g,
```
skywritergr: Can we add a comment here that says what the regex does. In a couple of months noone will know what this thing is doing.

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,366 @@
...
+      (matchedSubstring, parenthesizedSubmatch) =>
```
skywritergr: Is matchedSubstring used anywhere?

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,366 @@
...
+   * @return {Boolean}        [Valid popover options]
+   */
+  function isOptionsObject(object) {
+    return curry(looksLike)({
+      title: 'Popover title',
+      content: 'Popover content',
+    })(object);
+  }
+
+  function looksLike(a, b) {
```
skywritergr: I don't like variable names `a` and `b` can we make them more meaningful?

src/help/pop-over/pop-over.service.js
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  /**
+   * FP helper functions
+   */
+  function compose(...fns) {
+    return fns.reverse().reduce((fn1, fn2) => (...args) => fn2(fn1(...args)));
+  }
+
+  function curry(fn, arity = fn.length) {
```
skywritergr: This is slightly blowing my mind. A function, that returns a function, that returns a functions that returns the result of itself? What are we trying to achieve here?

demo/partials/tw-pop-over.html
mstaicu: The attribute, `data-content-html`, value should contain the popover template. I just removed `true` because if a value is present, it is used as a template for the popover, and since `true` is not a valid template and I wanted to leave the attribute as an example, I decided to leave it as an empty string.

src/help/pop-over/pop-over.directive.js
mstaicu: I will remove the commented out examples.

src/help/pop-over/pop-over.service.js
mstaicu: Apparently you're right. If we're not using the `babel-node`, which loads the polyfill to emulate a full ES2015+ environment, we cannot use new built-ins like `Promise` or `WeakMap`, static methods like `Array.from` or `Object.assign`, instance methods like `Array.prototype.includes`. I can't see the `babel-node` in the dependencies of the `styleguide-components`, thus I can assume that we don't load it. I will write the implementation without `Array.from`.

src/help/pop-over/pop-over.service.js
mstaicu: The `getPopoverTemplate` function has the appropriate comment which explains why we need to have that naming convention in the RegEX.

Each value in the template, which is equivalent to the keys of the binded object, is prefixed and suffixed with '__'. For example '__title__' corresponds to the value of the 'title' property on the binded object.

src/help/pop-over/pop-over.service.js
mstaicu: No, it's not, but I need it in my parameter list to get to `parenthesizedSubmatch`.

src/help/pop-over/pop-over.service.js
mstaicu: I'm using function composition to have a declarative approach instead of an imperative approach, thus making it more easily to remove a function from the composition instead of rewriting imperative procedural code.

Composes functions from right to left. Each function is expected to accept a single parameter. Its return value will be provided as an argument to the function standing to the left, and so on. The exception is the right-most argument which can accept multiple parameters, as it will provide the signature for the resulting composed function.

src/help/pop-over/pop-over.service.js
mstaicu: Yes we can, I was trying to do save Babel some work by writing minified code from the get-go. 🍾💥

src/help/pop-over/pop-over.service.js
mstaicu: I also added a more thorough explanation in the `.replace` part of the function.

src/help/pop-over/pop-over.service.js
mstaicu: The bulk of the code is made from pure functions that have a single responsibility and do `one thing`:tm:. The impure, procedural stuff is isolated in the following places: `showPopover`, `setPopoverPosition`, `hideCallback`, `registerGlobalEventListeners` and `unregisterGlobalEventListeners`. So it should be quiet easy to debug if the behaviour deviates from the expected etiquette.

src/help/pop-over/pop-over.directive.spec.js
```diff
@@ -0,0 +1,77 @@
...
+
+  describe('when we click on a popover trigger', function () {
+    beforeEach(function () {
+      element = getCompiledDirectiveElement($scope)[0];
+      element.click();
+    });
+
+    it('should append a popover on the body', function () {
+      var popover = document.querySelector('.popover');
+      var bodyChildren = [].slice.call($window.document.body.children);
```
StevePole: This is cryptic, can you make it less clever and more readable?

src/help/pop-over/pop-over.service.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      service.showPopover(elementWithPopover, popoverOptions);
+    });
+
+    afterEach(function() {
+      $window.document.body.removeChild(document.querySelector('.has-popover'));
+      $window.document.body.removeChild(document.querySelector('.popover'));
+    });
+
+    it('should append the popover to the body element', function () {
+      var popover = document.querySelector('.popover');
```
StevePole: We can move some of this test setup code into the beforeEach as it is shared, this will reduce the amount of code and simplify the expectationts.

src/help/pop-over/pop-over.service.js
StevePole: The questions George raises is whether this is readable.  If it's hard for others to understand, it's hard to maintain, and easy to introduce bugs.

src/help/pop-over/pop-over.service.spec.js
```diff
@@ -9,13 +9,14 @@ describe('TwPopoverService', function() {
...
-      'info': 'More info',
+      'title': '<span>Transferwise</span>',
+      'content': '<span>Rules</span>',
+      'html': 'true',
+      'image': 'http://transferwise.com/logo.png',
       'template': " \
-        <div class='popover-title'>__title__</div> \
-        <div class='popover-content'>__content__</div> \
-        <div class='popover-info'>__info__</div> \
+        <h3 class='popover-title'></h3> \
```
StevePole: In the Bootstrap popover, don't you pass in the `.popover` wrapper?

src/help/pop-over/pop-over.service.js
```diff
@@ -486,9 +466,39 @@ export default function () {
...
+      const popoverElement = popoverContainer.querySelector(popoverElementSelector);
+      const popoverElementValue = options[optionKey];
+
+      popoverElement.innerHTML = '';
+
+      const insertMethod = shouldRenderHTML ?
+        'insertAdjacentHTML' :
+        'insertAdjacentText';
+
+      popoverElement[insertMethod]('beforeend', popoverElementValue);
```
StevePole: I don't understand what's happening here, can you make it more readable?

PR #130, opened 2017-09-26
ASIA-920: Add date format filter

timotheehub:
ASIA-920: Add date format filter


src/formatting/date-format/date-format.filter.js
```diff
@@ -0,0 +1,25 @@
+import angular from 'angular';
+import DateFormatService from './date-format.service.js';
+
+function DateFormatFilter(TwDateFormatService) {
+  return (date, locale, patternName, hasWeekday, hasDayTime, hasSuffix) => {
```
StevePole: I think we can be more restrictive about how dates are displayed, and limit the options a little.  We shouldn't aim to support everything we do today, because there's no reason for us to have so much inconsistency.

I'm tempted to split out dates with times to a separate filter, which will help, but I think we can go much further.  Ideally I'd like to get to a point where we only have locale as a parameter.  Something like this:

{{$ctrl.date | twDate:'en_GB' }}
{{$ctrl.date | twDateTime:'en_GB' }}

What are the barriers to this?

src/formatting/date-format/date-format.service.js
```diff
@@ -0,0 +1,107 @@
...
+        date.getHours(),
+        date.getMinutes(),
+        locale
+      );
+      result = `${result} ${dayTime}`;
+    }
+    return result;
+  };
+
+  function addSuffix(monthDate, locale) {
```
StevePole: I feel like this could be part of the main date service.  Do you know if the behaviour is consistent across languages (particularly in the teens)?  Could we separate this to an object:
```
postfixes: {
  'en': {
    '1': 'st',
    '2': 'nd',
    '3': 'rd',
    'default': 'th'
  }
}
```
Or something along those lines?

src/services/date/date.service.js
```diff
@@ -146,19 +171,67 @@ function DateService() {
...
+    return `${date}/${month}/${year}`;
+  };
+
+  this.getMonthDatePresentation = (monthName, date, locale) => {
+    const lang = getLanguageFromLocale(locale);
+    if (lang === 'ja') {
+      return `${monthName}${date}日`;
+    }
+
+    if (locale && locale.indexOf('US', locale.length - 2) !== -1) {
```
StevePole: This logic should be centralised to a 'isMonthBeforeDate' function, which I think we may already have.

src/services/date/date.service.js
```diff
@@ -146,19 +171,67 @@ function DateService() {
...
+    return `${date} ${monthName}`;
+  };
+
+  this.getMonthDateShortPresentation = (month, date, locale) => {
+    const lang = getLanguageFromLocale(locale);
+    if (lang === 'ja') {
+      return `${month}月${date}日`;
+    }
+
+    if (locale && locale.indexOf('US', locale.length - 2) !== -1) {
```
StevePole: isMonthBeforeDate

src/services/date/date.service.js
```diff
@@ -146,19 +171,67 @@ function DateService() {
...
 
-    if (locale.indexOf('US', locale.length - 2) !== -1) {
+    if (locale && locale.indexOf('US', locale.length - 2) !== -1) {
       return `${monthName} ${date}, ${year}`;
     }
 
     return `${date} ${monthName} ${year}`;
   };
 
+  this.getYearMonthDateShortPresentation = (year, month, date, locale) => {
```
StevePole: If we were more deliberate about the formats we choose, we could remove a lot of this code.

src/formatting/date-format/date-format.service.js
```diff
@@ -0,0 +1,107 @@
+import angular from 'angular';
+import DateService from '../../services/date/date.service.js';
+
+function DateFormatService() {
```
StevePole: Not 100% sure if this file should exist.  Perhaps some of it should be in the main DateService, the rest in the date-format.filter.js.  If we allow less customisability, there will be less code.

src/services/date/date.service.js
```diff
@@ -259,7 +345,28 @@ function DateService() {
     ja: [
       '月', '火', '水', '木', '金', '土', '日'
     ]
+  };
 
+  const ORDINALS_BY_LANGUAGE = {
```
StevePole: Could we make this DAY_ORDINALS, then add MONTH_ORDINALS and YEAR_ORDINALS.  Then for Japan, you can add `年` as the year ordinal.  For month you can add the `月` ordinal.  Then I think we can remove more `if (lang === 'ja')` logic from above.

src/services/date/date.service.js
```diff
@@ -152,13 +177,70 @@ function DateService() {
       return `${year}年${monthName}${date}日`;
```
StevePole: I think we can remove this using YEAR, MONTH and DAY ordinals.

src/services/date/date.service.js
StevePole: I'm also now questioning whether 'SUFFIX' may be a better name for these things 🤔

src/services/date/date.service.js
```diff
@@ -137,6 +137,31 @@ function DateService() {
...
+    }
+    if (minutes < 10) {
+      minutes = `0${minutes}`;
+    }
+
+    if (lang === 'ja') {
+      return `${hours}:${minutes}の`;
+    }
+
+    if (locale && locale.startsWith('en')) {
```
StevePole: why not `lang`?

src/services/date/date.service.js
```diff
@@ -137,6 +137,31 @@ function DateService() {
...
+      const ampm = hours >= 12 ? 'PM' : 'AM';
```
StevePole: Can we do these in lowercase?

src/services/date/date.service.js
```diff
@@ -137,6 +137,31 @@ function DateService() {
...
+    const lang = getLanguageFromLocale(locale);
+    if (hours < 10) {
+      hours = `0${hours}`;
+    }
+    if (minutes < 10) {
+      minutes = `0${minutes}`;
+    }
+
+    if (lang === 'ja') {
+      return `${hours}:${minutes}の`;
```
StevePole: Would prefer a getTimeSuffix(locale, hours) function that returns am/pm or the Japanese suffix as appropriate.

src/formatting/date-format/date-format.filter.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  beforeEach(module('tw.styleguide-components'));
+
+  beforeEach(inject(function($injector) {
+    $rootScope = $injector.get('$rootScope');
+    $compile = $injector.get('$compile');
+    $timeout = $injector.get('$timeout');
+    $scope = $rootScope.$new();
+  }));
+
+  describe('Given there is a locale', function() {
```
StevePole: 'Given the locale is en-GB'

src/formatting/date-format/date-format.filter.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      $element = getCompiledDirectiveElement($scope);
+      $scope.$apply();
+    });
+
+    it('should format the value', function() {
+      var textValue = $element.text().trim();
+      expect(textValue).toBe('5th November 2016');
+    });
+  });
+
```
StevePole: Add test for 'Given the locale is in Japanese'

src/formatting/date-format/date-format.filter.spec.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+describe('DateFormat filter, ', function() {
+  var $compile,
+      $rootScope,
+      $scope,
+      $element,
+      $timeout,
```
StevePole: Delete

src/formatting/date-format/date-format.filter.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      input;
+
+  beforeEach(module('tw.styleguide-components'));
+
+  beforeEach(inject(function($injector) {
+    $rootScope = $injector.get('$rootScope');
+    $compile = $injector.get('$compile');
+    $timeout = $injector.get('$timeout');
```
StevePole: Delete

src/services/date/date.service.js
timotheehub: Actually, this would be more complex because we cannot simply add am/pm but needs to change the hours.

src/services/date/date.service.js
```diff
@@ -152,13 +177,70 @@ function DateService() {
...
       return `${monthName} ${date}, ${year}`;
     }
 
     return `${date} ${monthName} ${year}`;
   };
 
+  this.getMonthDatePresentation = (monthName, date, locale) => {
+    const lang = getLanguageFromLocale(locale);
+    if (lang === 'ja') {
+      return `${monthName}${date}日`;
```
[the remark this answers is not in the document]
timotheehub: One difference is that there is no space for Japanese formats but I can add a variable for this.

src/services/date/date.service.js
timotheehub: YEAR_SUFFIXES was easy to add but MONTH_SUFFIXES is harder because a month is a name and we need a method that is able to return month names (i.e. January in English and 1月 in Japanese). It's why it's easier to consider "1月" as a string and not as "1" and a suffix.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "tw-styleguide-components",
-  "version": "2.0.9",
```
StevePole: New functionality is minor version, so 2.1.0

src/formatting/date-format/date-format.filter.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    return TwDateService.getDateFormat(date, locale);
+  };
+}
+
+DateFormatFilter.$inject = ['TwDateService'];
+
+export default angular
+  .module('tw.styleguide.formatting.date-format', [DateService])
```
StevePole: 'tw.styleguide.formatting.date'

src/formatting/date-format/date-time-format.filter.js
```diff
@@ -0,0 +1,24 @@
...
+      locale
+    );
+    return `${dateText} ${timeText}`;
+  };
+}
+
+DateTimeFormatFilter.$inject = ['TwDateService'];
+
+export default angular
+  .module('tw.styleguide.formatting.date-time-format', [DateService])
```
StevePole: 'tw.styleguide.formatting.date-time'

src/formatting/date-format/date-time-format.filter.js
```diff
@@ -0,0 +1,24 @@
...
+import DateService from '../../services/date/index.js';
+
+function DateTimeFormatFilter(TwDateService) {
+  return (date, locale) => {
+    if (!date) {
+      return date;
+    }
+
+    const dateText = TwDateService.getDateFormat(date, locale);
+    const timeText = TwDateService.getTimePresentation(
```
StevePole: Could you change these two functions to be called:
```
const dateText = TwDateService.getLocaleDateString(date, locale);
const timeText = TwDateService.getLocaleTimeString(date, locale);
```
I would prefer to be explicit that we're using locale interpretation of the date object.
For the time, I prefer to use the date object for a couple reasons: conformity of interface, and because out current interface would not allow us to use seconds in the future without refactoring.

PR #133, opened 2017-10-05
Fix issue with select on docs.

StevePole:
Updates a number of dependencies, including eslint and webpack dev server.


src/forms/select/select.html
```diff
@@ -3,7 +3,7 @@
     dropdown: !$ctrl.dropdownUp,
     dropup: $ctrl.dropdownUp
   }" aria-hidden="false">
-
+
```
dey-dey: extra space?

src/forms/select/select.controller.js
```diff
@@ -3,19 +3,27 @@ import angular from 'angular';
...
+    this.dom = TwDomService;
+
     this.element = $element[0];
+  }
+
+  $onInit() {
     this.button = this.element.getElementsByClassName('btn')[0];
     this.search = '';
 
-    this.dom = TwDomService;
```
dey-dey: 👍

PR #135, opened 2017-10-05
Add field specific errors and fix onRefreshRequirements

kstam:
This is to deal with:
https://github.com/transferwise/styleguide-components/issues/134


src/forms/fieldset/fieldset.controller.js
```diff
@@ -135,6 +134,13 @@ function getParamValuesFromModel(model, params) {
   return data;
 }
 
+function prepValidationMessages(field, validationMessages) {
+  const hasValidationMessages = angular.isObject(field.validationMessages);
```
StevePole: Could we do this natively?

PR #138, opened 2017-10-10
Definition lists, Support flat requirement

StevePole:
The old requirements format expects the controls to be in a nested ‘group’ within the field.  this allows us to flatten that structure.
```
"fields": [
      {
         "name": "Card number",
         "group": [
           {
             "key": "number",
             "type": "text",
             …
```
Becomes:
```
"fields": [
      {
         "name": "Card number",
         "key": "number",
         "type": "text",
          …
```
We retain support for the old format as well.

We also add a definition list component for displaying the models generated by a dynamic form.

image

image

image


src/forms/definition-list/definition-list.controller.js
```diff
@@ -0,0 +1,40 @@
...
+class DefinitionListController {
+  constructor($scope, TwRequirementsService) {
+    this.$scope = $scope;
+    this.RequirementsService = TwRequirementsService;
+  }
+
+  $onInit() {
+    this.RequirementsService.prepFields(this.fields, this.model);
+
+    this.$scope.$watch('$ctrl.fields', (newValue, oldValue) => {
```
skywritergr: While this is not wrong, we use $ctrl in all of our controllers and it might be confusing. I wouldn't mind if we changed it to not have it in there.

src/forms/definition-list/definition-list.html
```diff
@@ -0,0 +1,67 @@
...
+  <h4
+    ng-if="$ctrl.legend"
+    ng-bind="$ctrl.legend"
+    class="page-header p-t-3">
+  </h4>
+</div>
+<div class="row row-equal-height">
+  <div class="col-xs-12"
+    ng-class="{'col-sm-6': !$ctrl.narrow}"
+    ng-repeat="field in $ctrl.fields"
```
skywritergr: we should try and make an effort to add indexes in our ng-repeats. It helps for performance reasons.

src/forms/fieldset/fieldset.html
```diff
@@ -1,64 +1,110 @@
...
   <legend ng-if="$ctrl.legend">{{$ctrl.legend}}</legend>
   <div class="row row-equal-height">
-    <div ng-repeat="fieldGroup in $ctrl.fields" class="col-xs-12"
+    <div ng-repeat="field in $ctrl.fields" class="col-xs-12"
       ng-class="{
-        'col-sm-4': fieldGroup.width === 'sm',
-        'col-sm-6': fieldGroup.width === 'md' || fieldGroup.maxlength && fieldGroup.maxlength <= 10,
-        'col-sm-12': fieldGroup.width === 'lg' || !fieldGroup.maxlength || fieldGroup.maxlength > 10
+        'col-sm-4': !$ctrl.narrow  && field.width === 'sm',
+        'col-sm-6': !$ctrl.narrow  && (field.width === 'md' || field.maxlength && field.maxlength <= 10),
```
skywritergr: I have a general preference to put these things into function when they are more than 2 arguments long. But up to you.

src/services/requirements/requirements.service.js
```diff
@@ -0,0 +1,101 @@
...
+  this.prepFields = (fields, model, validationMessages) => {
+    if (!fields) {
+      return;
+    }
+    fields.forEach((field) => {
+      if (field.group) {
+        if (field.group.length) {
+          field.key = field.group[0].key;
+        }
+        field.group.forEach((fieldSection) => {
```
skywritergr: These nested for loops are not ideal. Why are we doing this? is there any way to flatten this and do 2 loops, one after the other, instead of two nested loops?

src/services/requirements/requirements.service.js
```diff
@@ -0,0 +1,101 @@
...
+    if (field.valuesAsync.params &&
+      field.valuesAsync.params.length) {
+      postData = this.getParamValuesFromModel(model, field.valuesAsync.params);
+    }
+
+    this.$http.post(field.valuesAsync.url, postData).then((response) => {
+      field.valuesAllowed = response.data;
+      this.prepValuesAllowed(field);
+    }).catch(() => {
+      // TODO - RETRY?
```
skywritergr: We should either do something here or throw an error.

src/services/requirements/requirements.service.js
StevePole: They're looping over nested things, performance characteristic is O(n).  They make the requirements easier to consume, and allow us to support older versions of the format.

demo/json/account-requirements.json
```diff
@@ -4,90 +4,78 @@
...
+        "displayFormat": null,
+        "example": "ABCDDE22",
+        "minLength": null,
+        "maxLength": null,
+        "validationRegexp": "(?i)[A-Z]{6}[A-Z\\d]{2}([A-Z\\d]{3})?",
+        "validationAsync": {
+          "url": "https://api.transferwise.com/v1/validators/bic",
+          "params": [
+            {
+              "key": "iban",
```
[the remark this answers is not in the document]
StevePole: This file is only used for the demo page.  The keys were copied from the real API, but I agree it's not ideal.  I'll change.

src/forms/definition-list/definition-list.controller.js
StevePole: I don't understand this comment...?

StevePole:
I agree on console.warn, that's more semantic, but it doesn't eliminate the need for the eslint ignore.

src/forms/fieldset/fieldset.html
StevePole: Think I'm going to leave it for now, probably going to remove the stuff around maxlength in future.

src/forms/definition-list/definition-list.controller.js
skywritergr: It might feel like we are going to use a variable from a controller or something. `$ctrl = this;` is something we do across the frontend apps all the time.

PR #141, opened 2017-10-17
Add IE 9, 10, 11 support to events

vahurh:
- replace all cases of creating a new Event with CustomEvent
- add a polyfill for CustomEvent for when it's not defined (e.g IE 9, 10, 11)


src/index.js
```diff
@@ -10,11 +10,14 @@ import Layout from './layout/';
 import Loading from './loading/';
 import Navigation from './navigation/';
 
+import initPolyfills from './polyfill';
```
StevePole: I'd swap this for `import './polyfill';`. then keep the polyfill a bit closer to the original, bring back the closure, and run the code, and make it run the code so you don't need to do `initPolyfills()` below.  We don't really want to do anything with polyfill Js, just to make sure it has run.  I'd also move this to be the first import.

StevePole: Could you also add the import to `form-components.js` and `form-validation.js`.  These module entry points are there to support consumers using the old builds.

PR #142, opened 2017-10-19
Add input type=hidden and hidden fieldset

vahurh:


src/forms/dynamic-form-control/form-control.html
```diff
@@ -122,4 +122,10 @@
       ng-focus="$ctrl.focus()"
       ng-blur="$ctrl.blur()"></tw-date>
   </div>
+  <input ng-switch-when="hidden"
+      name="{{$ctrl.name}}"
+      type="hidden"
+      class="form-control"
```
StevePole: Not needed as hidden.

src/forms/dynamic-form-control/form-control.html
```diff
@@ -122,4 +122,10 @@
...
+      ng-model="$ctrl.ngModel"
+      ng-model-options="{ allowInvalid: true }"/>
```
StevePole: Is this necessary?  I can't think of a reason why as there are no validation rules on the control.

vahurh: didn't really know what it did, should've taken a 'lets remove & test if it breaks smth'-approach. :)

PR #143, opened 2017-10-25
Format iso strings without times as UTC

StevePole:


src/services/date/date.service.js
```diff
@@ -85,16 +125,21 @@ function DateService() {
...
+      defaultDayName = DEFAULT_DAY_NAMES_BY_LANGUAGE[language][dayOfWeek];
     }
 
     const validLocale = getValidLocale(locale);
     const date = this.getUTCDateFromParts(2006, 0, dayOfWeek + 1); // 2006 started with a Sunday
 
-    return getLocalisedDateName(date, validLocale, { weekday: format });
+    const autoDayName = getLocalisedDateName(date, validLocale, { weekday: format });
+
+    if (defaultDayName && autoDayName.length > 5) {
```
timotheehub: Why do we check the length of autoDayName?

src/services/date/date.service.js
```diff
@@ -208,73 +258,133 @@ function DateService() {
...
   };
 
-  this.getLocaleFullDate = (date, locale) => {
-    const dateString = this.getYearMonthDatePresentation(
+  this.getLocaleNow = () => new Date();
+
+  this.getUTCNow = () => {
+    const now = new Date();
+    return this.getUTCDateFromParts(
+      this.getUTCFullYear(now),
```
timotheehub: why are we using this.getUTCFullYear(now) instead of now.getUTCFullYear()?

src/services/date/date.service.js
```diff
@@ -208,73 +258,133 @@ function DateService() {
...
 
-    if (hasWeekday) {
+    if (dayName) {
       if (lang === 'ja') {
-        return `${dateString} ${timeName} (${weekdayName})`.trim();
+        return `${dateString} ${timeName} (${dayName})`.trim();
+      }
+      // For longer dates use a comma
+      if (monthName || yearName) {
+        dateString = `${dayName}, ${dateString}`.trim();
```
timotheehub: This comma works in some languages like English and German but might not work in all the languages (e.g. French). Should we add it only for english?

src/formatting/date-format/date-format.filter.js
```diff
@@ -4,16 +4,24 @@ function DateFormatFilter(TwDateService) {
     if (!date) {
       return date;
     }
+    let utcDate = date;
```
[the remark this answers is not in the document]
StevePole: Fair point, will improve 👍

src/formatting/date-format/date-format.filter.js
```diff
@@ -4,16 +4,24 @@ function DateFormatFilter(TwDateService) {
...
-    if (format === 'long') {
-      return TwDateService.getLocaleFullDate(date, locale);
+        return TwDateService.getUTCDateString(utcDate, locale, format);
+      }
     }
 
-    return TwDateService.getLocaleDateString(date, locale, format === 'short');
+    if (!utcDate) { return date; }
+    // Use locale timezone
+    const ret = TwDateService.getLocaleDateString(utcDate, locale, format);
```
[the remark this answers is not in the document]
StevePole: Yeah, I had a console log in there previously and forgot to tidy.

src/services/date/date.service.js
StevePole: Some browsers do no have the ability to shorten the name, this lets us shorten it if that's the case.

src/services/date/date.service.js
StevePole: Good point, will change.

src/services/date/date.service.js
StevePole: They still have punctuation right? :-)  I don't think it can hurt too badly, this only comes in if we're generating a long name.  Including it here allowed me to remove `getLocaleFullDate` which previously had this behaviour.

src/services/date/date.service.js
```diff
@@ -74,6 +74,46 @@ function DateService() {
...
+
+  this.isIsoStringValid = (isoDate) => {
+    const dateSection = '[0-9]{4}-[0-9]{2}-[0-9]{2}';
+    const timeSection = 'T[0-9]{2}:[0-9]{2}:[0-9]{2}';
+    const zoneSection = '(Z|[+,-][0-9]{2}(:[0-9]{2})?)';
+    const regex = new RegExp(`^${dateSection}(${timeSection}${zoneSection})?$`);
+    return regex.test(isoDate);
+  };
+
+  this.getUTCDateFromIso = (isoDate) => {
```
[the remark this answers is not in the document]
StevePole: Good spot, I've switched back to using it as it's safer than using the Date constructor.

src/services/date/date.service.js
timotheehub: Haha, yes. I let you decide. A comma between the weekday and the rest of the date would look odd to native French speakers but it is just a detail.

PR #144, opened 2017-11-09
Adds help images for fieldset and form control

Tankenstein:
By adding `helpImage` to a field you render an image above the field with that url, except for upload fields where it replaces the icon.

image


StevePole:
image

:-o

Need to choose some spacing that keeps the visual rhythm of a form, guessing `.m-b-3`

src/forms/dynamic-form-control/form-control.component.js
```diff
@@ -19,6 +19,7 @@ const FormControl = {
     uploadAccept: '@',
     uploadIcon: '@',
     uploadTooLargeMessage: '@',
+    uploadHelpImage: '@',
```
StevePole: Let's just make this helpImage as it can be used by other types of components.

PR #145, opened 2017-11-10
Add currency formatter.  Improve number formatter in Safari.

StevePole:


src/formatting/currency-format/currency-format.filter.js
```diff
@@ -0,0 +1,33 @@
+
```
skywritergr: Empty line

PR #147, opened 2017-11-22
Correct 'html' dataset attribute to 'content-html'

mstaicu:


dist/js/styleguide-components.js
```diff
@@ -5459,8 +5459,8 @@ function getElementOptions(element) {
   if (element.dataset.template) {
     options.template = element.dataset.template;
   }
-  if (element.dataset.html) {
-    options.html = element.dataset.html === 'true';
+  if (element.dataset.contentHtml) {
+    options.contentHtml = element.dataset.contentHtml === 'true';
```
skywritergr: I would extract this to a variable or add parentheses around it.

PR #148, opened 2017-11-29
Add 'container' option to the popover service and directive

mstaicu:


dist/js/styleguide-components.js
```diff
@@ -5465,6 +5465,9 @@ function getElementOptions(element) {
   if (element.dataset.image) {
     options.image = element.dataset.image;
   }
+  if (element.dataset.container) {
```
skywritergr: Should we use a `switch` statement here for all those ifs?

dist/js/styleguide-components.js
```diff
@@ -5672,11 +5676,13 @@ function PopoverService() {
    * @return {String}           [Popover's new placement]
    */
   function checkPopoverPlacement(placement) {
+    var popoverContainerElement = getPopoverParentElement(elementPopoverOptions);
+    var coordinateComputeFunction = popoverContainerElement === BODY ? getBoundingOffset : getParentOffset;
```
skywritergr: I don't see the `getBoundingOffset` and `getParentOffset` in the scope of this function. Where are they coming from?

mstaicu: We're computing the coordinates relative to the viewport or relative to the first offset parent based on the option that we pass via the `container` property. The `getBoundingOffset` and `getParentOffset` are functions that do just that, and we're deciding which coordinate function we're using based on the value passed as the `container`. Those functions are not scoped in the same function, they are declared as part of the service.

dist/js/styleguide-components.js
mstaicu: This is turning into a spaghetti code, I'm thinking of something right now.

src/help/pop-over/pop-over.directive.spec.js
```diff
@@ -67,11 +67,12 @@ describe('Popover directive', function() {
...
         <a tw-pop-over \
           data-original-title='<span>Popover title</span>' \
-          data-content='Full description copy, explaining in more detail'\
-          data-image='http://www.transferwise.com/logo.png'\
+          data-content='Full description copy, explaining in more detail' \
+          data-container='body' \
+          data-image='http://www.transferwise.com/logo.png' \
           data-template='<div class=\"popover\"> \
                            <h3 class=\"popover-title\"></h3> \
-                           <img class=\"popover-image\" alt=\"Popover content\" /> \
```
StevePole: Why did you remove the closing tag?

mstaicu: The  tag represents what is known as a void element so called because it can't have any contents. Therefore there is no syntactic reason why it should need to be closed in HTML. If the element is one of the void elements, in our case ``, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements.

StevePole: Does htmllint agree?  If it passes that's fine.

PR #149, opened 2017-11-29
Add phone number component

lapostoj:
image


src/forms/phone-number/index.js
```diff
@@ -0,0 +1,6 @@
+import angular from 'angular';
+import PhoneNumberControl from './phone-number.component.js';
+
+export default angular
+  .module('tw.styleguide.forms.phone-number', [])
+  .component('twPhoneNumber', PhoneNumberControl).name;
```
lapostoj: this is the same name the current one has in frontend-apps.
So either we need to go update the name there before putting this one in the wild or find another name to this one.
Opinions on what is best?

StevePole:
Awesome, thanks for this!  Reviewing from phone, so will leave comments here:

Could we make the countries optional?  There is a finite set that could be provided by a service, in most cases I think you would be able to pick any of them.  You could supply an override to limit the set.  In either case, could we allow you to pass just the country code, and the component will know the rest e.g. the +prefix?  Just trying to make it simpler to use.  We’ll need to add this to the dynamic form, so if the consumer needs to provide it we’ll have to have the dynamic form do it somehow, a bit messy...

Ideally I think of these as direct replacements for the native input[type=tel] and they shouldn’t be harder to use.  In almost all cases the name of the component also mirrors the input type (select, radio, checkbox etc), although I screwed up file (twUpload), we could go twTel, but perhaps twTelephone is better? 

Not sure about the transclusion approach to validation, I know it worked like this somewhere else, but I’ve generally moved away from it, preferring to make sure the validity of the ngModelController on the component is set correctly, that way these components are more interchangeable with the native HTML variants.

There is an optional parameter on twSelect ‘hide-note’ that will hide the extra info of the selected value when the control is closes.

You have an Fdescribe in your tests

lapostoj:
So first, I added the `hide-note` and renamed to tw-telephone, I think it makes sense. (and removed the fdescribe obviously)...

For the other points they could need a bit more discussion (at different levels).

About the country list, I think it makes sense to have reasonable set inside the component in case nothing is provided yes. The only issue I see is about the translation of the country names. But here either we can go with english (if it's a fallback) or just don't have the names but only the phone codes in this case.
Mapping the country code to the phone code should also be possible I guess.
One side issue with the countries is when there are several countries with the same phone code...
When you select it there is no issue, but then if you are shown the component filled with the value you entered previously it would select the first country with the phone code you have... (for example Guernsey is before United Kingdom in the list). That is a problem we already have with the current component and which make me wonder if the country names should even belong here...
But on the other hand they can be nice to search the value for users...

About the transclusion for the validation what I like is that it come almost for free here while otherwise there is a bit of boilerplate to bind the validation correctly which might not even be that obvious. While the results are basically the same if I'm right... Then it adds a small condition to the use of the component yes... But since the error messages have to be passed anyway is it really much harder to pass them next to the component or inside it? 🤔 

Another note about "that way these components are more interchangeable with the native HTML variants."... It might not have been obvious on phone but for now I have the `ngModel` as a one way binding and a callback to propagate changes happening to it. Benefit is again a bit less boilerplate to update the internal model from the binding. But with the idea you seem to have in mind for the components it seems it might not be the best...

On a slightly different note not sure if trying to be interchangeable with native HTML should apply too much here... Since the input[type=tel] have very little native logic and here we even split it into 2 inputs one for the prefix and one for the rest of the number, it seems we are diverging already quite a bit... (but we could still try to limit the divergence if we want yes).

StevePole:
On being interchangeable with the native, I don't mind if the implementation is different, in fact that's kind of the point, just that it's interface is as compatible as possible with native.  So, if I'm currently using:
```
<input type="tel" ng-model="..." ng-required="..." />
```
I would like to be able to replace it with:
```
<tw-telephone  ng-model="..." ng-required="..." />
```
The UX will be different (better), but the code is the same.

StevePole:
On country names vs country codes.  Given the problem with pre-setting I think you may be right, maybe we should remove.  Or we could consider +44 Uk, gurnsey etc.  Also a problem for translation.  Slight aside, but when you split, how do you know whether to take 1 digit or 2?  E.g. US is '+1', but are there any other countries like '+17'?

StevePole:
If we went for number I think it would still be cool to use the 'searchable' feature of tw-select so that you could search by country name, even if only English.

StevePole:
On model binding, happy to keep the events, but I would like to keep it 2 way for consistency with the other components.  Once you start looking at tw-field it get's pretty weird if one component has to handled differently.

StevePole:
The validation without transclusion is actually not to difficult you just need to use (from memory) $ngModelController.$setViewValue(combinedValue) for the ngModelController on the component, which makes it run through the validators that are declared on the same element as the ngModel.

Like this: https://github.com/transferwise/styleguide-components/blob/master/src/forms/dynamic-form-control/form-control.controller.js#L17

PR #150, opened 2017-12-01
Add a LocaleService & use by default in formatters.  Add formatting components.

StevePole:
Add formatting components for number and currency formatters.
Refactor formatting module includes.


src/formatting/currency-format/currency-format.spec.js
```diff
@@ -6,26 +6,30 @@ describe('CurrencyFormat filter, ', function() {
...
       $scope.currency = 'GBP';
+      spyOn(LocaleService, 'get').and.returnValue('en-GB');
       $element = getCompiledDirectiveElement($scope);
       textValue = $element.text().trim();
     });
 
-    it('should default to en-GB format', function() {
+    it('should use the locale from the locale service', function() {
       if (isNumberLocaleSupported()) {
+        expect(LocaleService.get).toHaveBeenCalled();
```
lapostoj: Should that really be here? Because it all happens in the numberFormatFilter right? So we should only test it there I think

src/formatting/text-format/undo-stack.service.js
```diff
@@ -1,5 +1,3 @@
-import angular from 'angular';
-
 function TwUndoStackFactory() {
```
lapostoj: Does it need the `Tw` prefix here? (given that for other services you didn't put it)

src/services/locale/locale.service.js
```diff
@@ -0,0 +1,29 @@
+function LocaleService() {
+  this.locale = 'en-GB';
+  this.regex = /^[a-z]{2}(-[A-Z][A-Z])?$/;
+  this.lowerCaseCountry = /^[a-z]{2}(-[a-z][a-z])?$/;
+
+  this.get = () => this.locale;
```
lapostoj: Maybe make the name a bit more explicit? `LocaleService.get` doesn't seem to obvious about what you are getting.
Plus it would be a bit more future proof in case the service is extended later.

src/formatting/currency-format/currency-format.spec.js
StevePole: I did consider that, actually not the biggest fan of this test as it's quite coupled to the implementation, but I felt this this was more explicit.  I don't really care how the result is achieved, maybe the number format will get the locale, maybe this will do it and pass it into the number formatter, the important thing is that we fetch it.  If I were to mock the number formatter, the logic for that could change and it would no longer get the locale.  However, I think the real issue is the neither actually test the correct thing.  The test should set the locale, and then just see if we're using the right thing in the output.

src/formatting/text-format/undo-stack.service.js
StevePole: Not really part of this PR, but I'll indulge you!

lapostoj:
Would also need a version bump but I guess it's planned 🙂

PR #153, opened 2017-12-13
Local environment fixes potentially different locale

peter-czuczor:
There were two main issues with tests running on my local environement (macOS Sierra 10.12.6, Hungarian local added to my locals):

1) Running with `Chrome browser` and `karma-chrome-launcher` plugin. 
This is the result for me:

```
  CurrencyFormat filter,
    when no locale supplied
      ✖ should use the locale from the locale service
        Chrome 63.0.3239 (Mac OS X 10.12.6)
      Expected '1 234,56 GBP' to equal '1 234,56 GBP'.
          at UserContext.<anonymous> (/Users/peter.czuczor/Projects/styleguide-components/src/formatting/currency-format/currency-format.spec.js:36:27)
```
So you could not see any differences right? 
So am I, but checking it more carefully this can be seen:

```
' '.charCodeAt(0)
160
' '.charCodeAt(0)
32
```

Produced different whitespace character. It seems this one also depends on the local, so for further issues I would just replace these whitespaces with the same one from the element result with that code part:
```
      $element = getCompiledDirectiveElement($scope);
      textValue = $element.text().trim().replace(/\s/g,' ');
```

2) Running with `PhantomJS browser` and `karma-phantomjs-launcher` plugin.
These are the errors:

```
SUMMARY:
✔ 555 tests completed
ℹ 12 tests skipped
✖ 11 tests failed

FAILED TESTS:
  DateFormat filter,
    Given the locale is en-GB
      and a short format is requested
        ✖ should shorten the formatted value
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '5th .No 2016' to be '5th Nov 2016'.
        /Users/peter.czuczor/Projects/styleguide-components/src/formatting/date-format/date-format.filter.spec.js:89:31

  DateLookup,
    given the component is initialised
      when locale attribute is passed
        ✖ should populate vm.months with default English months
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '.January.' to be 'January'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:921:45
        forEach@[native code]
        expectDateMonthsToBeDefault@/Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:920:27
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:226:38

      when shortDate==true
        ✖ should show short month name
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '10th .Ja 2000' to be '10th Jan 2000'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:239:71

    when the component is already initialised
      given locale changes
        ✖ should populate $ctrl.months with default English months
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '.January.' to be 'January'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:921:45
        forEach@[native code]
        expectDateMonthsToBeDefault@/Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:920:27
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:515:40

        ✖ should show day before month if locale not US
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '1 .January. 2000' to equal '1 January 2000'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date-lookup/date-lookup.spec.js:525:76

  Date
    init
      when locale attribute is passed
        ✖ should populate $ctrl.months with default English months
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '.January.' to be 'January'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:832:49
        forEach@[native code]
        expectDateMonthsToBeDefault@/Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:830:27
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:295:38

      when incorrect locale  passed
        ✖ should populate $ctrl.dateMonths with default English months
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '.January.' to be 'January'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:832:49
        forEach@[native code]
        expectDateMonthsToBeDefault@/Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:830:27
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:320:36

    watchers on shared $scope
      locale
        ✖ should populate $ctrl.dateMonths with default English months
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '.January.' to be 'January'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:832:49
        forEach@[native code]
        expectDateMonthsToBeDefault@/Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:830:27
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/date/date.spec.js:651:38

  Definition list
    when given a date field
      ✖ should display the formatted date value
        PhantomJS 2.1.1 (Mac OS X 0.0.0)
      Expected '1 .January. 2000' to be '1 January 2000'.
      /Users/peter.czuczor/Projects/styleguide-components/src/forms/definition-list/definition-list.spec.js:81:49

    when given requirements with nested field groups
      when given a date field
        ✖ should display the formatted date value
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected '1 .January. 2000' to be '1 January 2000'.
        /Users/peter.czuczor/Projects/styleguide-components/src/forms/definition-list/definition-list.spec.js:233:51

  TwDateService test
    when English locale
       and short format
        ✖ should provide the correct month names
          PhantomJS 2.1.1 (Mac OS X 0.0.0)
        Expected $[0] = '.Ja' to equal 'Jan'.
        Expected $[1] = '.Fe' to equal 'Feb'.
        Expected $[2] = '.Ma' to equal 'Mar'.
        Expected $[3] = '.Ap' to equal 'Apr'.
        Expected $[4] = '.Ma' to equal 'May'.
        Expected $[5] = '.Ju' to equal 'June'.
        Expected $[6] = '.Ju' to equal 'July'.
        Expected $[7] = '.Au' to equal 'Aug'.
        Expected $[8] = '.Se' to equal 'Sep'.
        Expected $[9] = '.Oc' to equal 'Oct'.
        Expected $[10] = '.No' to equal 'Nov'.
        Expected $[11] = '.De' to equal 'Dec'.
        /Users/peter.czuczor/Projects/styleguide-components/src/services/date/date.spec.js:27:35

npm ERR! Test failed.  See above for more details.
```

The common point in all failing tests, that everywhere there is an extra point in the result.

So for example this is happening, because this method call:

`date.toLocaleDateString(locale, formattingObject);`

Gives back:

`'2000. November 15.'`

And so it is ignoring the `formatting object`, which should be:

`Object{month: 'long'}`

So should give back only month because of the formattingObject setup, but give back the whole Date.
And when we clearing the numbers and whitespaces from the date, we end up with:

`.November`

This way, even if we only check the first 3 characters we will end up with a `.` prefix always, and every date will contain this point `.`
The fix is to not only replace numbers, and whitespaces, but points, commas, semicolons also:
```
      // strip out any numbers, spaces and commas in case browser (cough...Safari)
      // doesn't respect format
      //Old solution: name = name.replace(/[0-9]|\s|,/g, '');
      name = name.replace(/[0-9]|[;,.]|\s|,/g, '');
```

+1) Updated `.gitignore` with gradle related files.

Thanks for reading it until end.
And just text me and ask, if some part is not understandable of this.


.gitignore
```diff
@@ -15,3 +15,7 @@ newrelic_agent.log
 .gradle/
 /gh-pages/
 .vscode/
+
+gradle/
```
coolsebz: this repository normally shouldn't have these files, why is this needed?

StevePole:
@coolsebz the challenge is this relies on a method on the browsers date object for formatting.  The only thing to do is mock out the date object.  Which I may do, but it also makes the tests a little pointless.

StevePole:
In some ways better to mock it, as this could have affected customers and we would not have known if we mocked the object...

src/formatting/currency-format/currency-format.spec.js
```diff
@@ -24,7 +24,7 @@ describe('CurrencyFormat filter, ', function() {
       $scope.currency = 'GBP';
       LocaleService.setCurrent('fr-FR');
       $element = getCompiledDirectiveElement($scope);
-      textValue = $element.text().trim();
+      textValue = $element.text().trim().replace(/\s/g,' ');
```
StevePole: Should this be in the test, or in the code?  I feel like the result will be more standard if it's in the code...

.gitignore
peter-czuczor: Maybe I messed something or my Intellij IDEA messed something ?

My thoughts:
It was imported and possiblz these directories were created during the import phase.
Do you want to kill, wipe this thing from the gitignore? 
Or can I leave it there for further IntelliJ users?

peter-czuczor:
Thanks Seb and Steve. Awesome feedbacks.
I agree to not mock date and or currency or other input based objects we have more coverage this way and can see issues more easy, but different browser engine and so implementation can mess up our method outputs and so tests :(

src/formatting/currency-format/currency-format.spec.js
peter-czuczor: I am really not sure, really.
The trim was in the test also, that is why I thought another string manipulation/replace can have a nice place here too, near to the trim().

The related code part:
```
  function getCompiledDirectiveElement(scope, template) {
    if (!template) {
      template = " \
        <span>{{ value | twCurrencyFormat : currency : locale }}</span>";
    }

    var element = angular.element(template);
    var compiledElement = $compile(element)(scope);
    scope.$digest();
    return compiledElement;
  }
```

Should I add after the `var compiledElement` part my regexp (`replace(/\s/g,' ')`)?
What do you think @StevePole?

StevePole: Thinking this through, the reason the trim is in the test, is because we don't care if there's whitespace in the implementation.  So from a behavioural point of view the following are equivalent:
```
<p class="one">
  My Text
</p>
```
and
`<p class="two">My Text</p>`
but... 
`$('.one').text() !== $('.two').text()`
I don't actually care how that template was written, I just care that the user sees the right thing, so it's fine (correct in fact) for the test to `.trim()`.

With the internal whitespace character, this is part of the value, and (technically) makes them different values, which is why I feel it might make more sense in the code.

peter-czuczor:
I will close this for now to ease my mind...  Sorry guys I did not have the capacity and the proper knowledge back then to properly close/merge this PR so it was pending for a while! And now it is just probably not relevant anymore.

I think the whole repository is changed much.

The whole point of this fix would be to ignore developer's local specific setups, like locale, time zone, and so do not make the tests fail because of that.

PR #154, opened 2017-12-15
Emit an event when the user interacts with the popover

mstaicu:


src/help/pop-over/pop-over.service.js
```diff
@@ -487,6 +495,10 @@ function PopoverService() {
    * @return {String} [CSS class]
    */
   function hidePopover() {
+    $rootScope.$emit('promotion:popoverCloseEvent', elementPopoverOptions);
```
StevePole: Could we change it to 3 events:
```
promotion:open
promotion:close
promotion:click
```
It's implicitly an event, and popover is just an implementation detail.  It would be good to have open and close for completeness, even if we're not using it right away.  I could imagine in the future using events like that to close other overlays (modals, etc).

PR #158, opened 2018-01-10
Fix fieldset docs.  Fix valuesAsync on fields

StevePole:


dist/js/styleguide-components.js
```diff
@@ -3802,6 +3807,10 @@ var FieldController = function () {
   return FieldController;
 }();
 
+function copyJSON(obj) {
+  return JSON.parse(JSON.stringify(obj));
```
skywritergr: I don't understand what's the purpose of this.

PR #159, opened 2018-01-11
Fix upload async failure

tomasqueirogabr:
Fixing two bugs:

1. A `.` was appearing in month name when running `npm test`, failing 11 tests. `date.toLocaleDateString` was returning `2000. January 15.`, so extracting the month would result in `.Ja` or `.January.`. There was already a regex to remove numbers and commas, but not dots, so the regex was updated to cover this scenario.
2. When trying to upload a file bigger than `maxSize` attribute of `tw-upload` was resulting in this error: `this.asyncFailure is not a function`. `asyncFailure` was moved out of the `UploadController` class, so it should be called without `this` and passing `this` as the last parameter.

screen shot 2018-01-11 at 16 17 07


src/forms/upload/index.js
```diff
@@ -1,9 +1,11 @@
 import angular from 'angular';
 
+import UploadController from './upload.controller.js';
 import Upload from './upload.component.js';
 import FileInput from './file-input.directive.js';
 
 export default angular
   .module('tw.styleguide.forms.upload', [])
+  .controller('UploadController', UploadController)
```
tomasqueirogabr: Needed to unit test `UploadController`

src/forms/upload/upload.spec.js
```diff
@@ -13,10 +15,12 @@ describe('Upload', function() {
 
   beforeEach(module('tw.styleguide.forms'));
 
-  beforeEach(inject(function($injector) {
+  beforeEach(inject(function($injector, _$controller_) {
     $rootScope = $injector.get('$rootScope');
     $compile = $injector.get('$compile');
     $scope = $rootScope.$new();
+    $timeout = $injector.get('$timeout');
+    $controller = _$controller_;
```
skywritergr: This is not the convention used here. Use the `$injector.get` format.

src/forms/upload/upload.spec.js
```diff
@@ -39,6 +43,28 @@ describe('Upload', function() {
...
+        $scope: $scope,
+        $transclude: function() {},
+        $q: {},
+        $attrs: {}
+      });
+      ctrl.reset();
+      ctrl.maxSize = 1;
+      var file = { size: 2 };
+
+      expect(ctrl.isError).toBe(false);
```
skywritergr: So wait, are you saying that you expect no error to be thrown when a big file is uploaded?

src/forms/upload/index.js
skywritergr: Yeah but now you are exposing it for no reason. If you check angular you can retrieve the controller via the component itself.

https://puigcerber.com/2016/02/07/how-to-test-angular-1-5-components/

```
beforeEach(function() {
  controller = element.controller('myComponent');
});
```

src/forms/upload/upload.spec.js
tomasqueirogabr: Here I expect to be no errors, since no file was uploaded yet. The line below `ctrl.fileDropped(file);` will "upload" the file, and then, after 3s, I except to have a error. `$timeout.flush(3001);` `expect(ctrl.isError).toBe(true);`

src/forms/upload/index.js
andrastothtw: We tried but somehow it just returned the template.

src/forms/upload/upload.spec.js
andrastothtw: I would add a comment above the `.toBe(false)` to clarify it does not have an error yet, because of a 3s timeout.

src/forms/upload/upload.spec.js
```diff
@@ -39,6 +43,28 @@ describe('Upload', function() {
     });
   });
 
+  describe('Failure state', function() {
+    it('when a big file is sent', function() {
+      var ctrl = $controller('UploadController', {
```
oliverviljamaa: Do we absolutely _need_ to test the controller? Does "failure state" never show to the customer? If it does, isn't that what we ultimately want to test, not the implementation details?

src/forms/upload/index.js
skywritergr: We use that convention in different parts. If i remember correct it should be in navigation too.

andrastothtw: We made a mistake here. I would say that people who design APIs that uses strings to access programmatic values should be expelled from the community.

src/forms/upload/upload.spec.js
tomasqueirogabr: I just changed to be as the example you posted above

src/forms/upload/upload.spec.js
tomasqueirogabr: I changed to retrieve the controller via the component, simplifying the code. How would you do to simulate the file upload without having the controller? I did not figure it out :(

PR #163, opened 2018-01-31
Transclude the transfer detail modal outside the expanded transclusion.

mstaicu:
This action is needed mainly because we apply CSS transformations on the parent of the details modal. Elements with transforms act as a containing block for fixed positioned descendants which render the desired behaviour in a different way.


StevePole:
Would it be possible to only apply the transforms at certain screen widths?  We definitely need them for mobile as the expanded section slides above the page, but perhaps not on desktop.  Of course, that would mean a different solution for the modal is needed on mobile...

mstaicu:
If you use the modal component within the `<expanded>` transclusion, it will not work on mobile, because the `.list-group-slide-out .collapse` is using an animation which applies a CSS `transform` on small and extra small screen devices. From what I've tried, there are only two solutions to this:

1. Use offsets instead of transforms, a solution which I clearly know it will get a 👎 
2. Create a new transclusion that is a sibling to the `expanded` transclusion, which won't apply CSS transforms and cause the `position: fixed;` modal component to be contained within containing block.

I think the second method is the cleanest, but the new transcluded element should be as generic as possible as @StevePole mentioned.

PR #164, opened 2018-02-01
Add tw-telephone to dynamic forms & update behaviour

StevePole:


src/forms/telephone/countries.json
```diff
@@ -0,0 +1,1263 @@
+[{
+	"name": "Afghanistan",
+	"iso2": "AF",
```
lapostoj: should we remove the iso2 from there if we don't use them?

src/forms/telephone/countries.json
```diff
@@ -0,0 +1,1263 @@
...
+	"name": "United Arab Emirates",
+	"iso2": "AE",
+	"iso3": "ARE",
+	"phone": "+971"
+}, {
+	"name": "United Kingdom",
+	"iso2": "GB",
+	"iso3": "GBR",
+	"phone": "+44",
+	"phoneFormat": "**** *** ***"
```
lapostoj: planning to have some formatting with that?

src/forms/telephone/countries.json
StevePole: I'm thinking about migrating this to a countries service, rather than having it specifically designed for this use case.  In which case I would want it.

src/forms/dynamic-form-control/form-control.html
```diff
@@ -90,6 +90,22 @@
     ></tw-date>
   </div>
 
+  <div ng-switch-when="tel">
```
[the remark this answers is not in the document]
StevePole: Well, I'm a bit torn.  From an API point of view, I like the idea of control type matching the native values, rather than our components.  Which means `<input type="tel" />`.  But it is also a little confusing.

src/forms/telephone/telephone.controller.js
```diff
@@ -1,82 +1,176 @@
...
+    this.onValueChange(
+      this.prefix,
+      // This shouldn't be necessary as $parser did it already, but otherwise test fails...
+      this.suffix && this.suffix.replace(this.charactersToRemove, '')
+    );
+  }
+
+  onValueChange(prefix, suffix) {
+    let combined;
+    // TODO safer to rely on validity in case we change to allowInvalid
```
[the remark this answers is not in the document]
StevePole: As of latest changes, no.  I'll remove.

src/forms/telephone/telephone.html
```diff
@@ -0,0 +1,34 @@
...
+      ng-model="$ctrl.suffix"
+      ng-model-options="{ allowInvalid: true }"
+      ng-change="$ctrl.onSuffixChange()"
+      ng-disabled="$ctrl.ngDisabled"
+      ng-focus="$ctrl.onFocus()"
+      ng-blur="$ctrl.onBlur()" />
+    <div ng-transclude class="error-messages"></div>
+  </div>
+</div>
+<!--
```
lapostoj: shouldn't these be put back in? 🤔 or otherwise deleted?

src/forms/telephone/telephone.spec.js
```diff
@@ -0,0 +1,393 @@
...
+      });
+      it('should set control values correctly', function() {
+        expect(getSelectValue(select)).toEqual("+44");
+        expect(input.value).toBe('123456789');
+      });
+      it('should leave number model as it was defined', function() {
+        expect($scope.ngModel).toBe('+44123456789');
+      });
+      it('should trigger the change handler', function() {
+        setNumberUsingControls(select, input, "+33", '09876543');
```
lapostoj: is this test in the right place? it doesn't really have to do with the fact that an model is initially passed, does it?

StevePole: Fair point. Will move.

StevePole: Actually, removing as tested by changes to the individual controls.

src/forms/telephone/telephone.controller.js
```diff
@@ -0,0 +1,201 @@
...
+
+  $onInit() {
+    this.callingCodes = codesToOptions(countryList);
+
+    this.formGroup = this.DomService.getClosestParentByClassName(
+      this.$element[0],
+      'form-group'
+    );
+
+    this.charactersPermitted = /^[0-9\s.-]*$/;
```
lapostoj: if we remove it from the template, this one becomes unused so we could remove as well.
I guess now it's replaced with the modelPattern you validate.

StevePole: Another good point.

PR #165, opened 2018-02-22
Add textarea to dynamic control

StevePole:
Refactor tests.


src/forms/dynamic-form-control/form-control.html
```diff
@@ -142,6 +142,22 @@
     ng-minlength="$ctrl.ngMinlength"
     ng-maxlength="$ctrl.ngMaxlength" />
 
+  <textarea ng-switch-when="textarea"
```
vahurh: Should we specify a default set of rows for the textarea? And I think we should allow setting a custom rows properties too.

StevePole: By default the height matches two controls.  

image

I'm reluctant to add direct control over height, but I'd consider if we need a 'small, medium or large', which is a bit more semantic, and leave the implementation up to the client.  However, I'm still not sure it's necessary...

PR #176, opened 2018-02-28
Fix automatic deploy of demo

lapostoj:


circle.yml
```diff
@@ -19,15 +19,15 @@ deployment:
      - npm run copy-files-to-dist
      - npm run build
      - npm run deploy
- releases:
-   branch: releases
-   commands:
-     - npm run release-to-github
-     - npm publish
      - git clone git@github.com:transferwise/styleguide-components.git gh-pages
```
lapostoj: this needs to happen on master as we don't copy the demo folder to the release branch (and I don't think we should)

PR #182, opened 2018-03-14
TW-Select not initialising correctly with ui-router navigation.

wilsonjs:
(@stevepole updated as old description was out of date)
Solved by updating to use one way binding for options and $onChanges over $scope.watch


StevePole:
Solved by updating to use one way binding for options and $onChanges over $scope.watch.

lapostoj:
I was wondering of the backward compatibility as well. I *think* in this case it should be ok, but would not be 100% confident either

StevePole:
I'm pretty confident, tests cover a large range of behaviours.  Logically I feel the code is doing what should be expected, and in a clearer way than before..

PR #183, opened 2018-03-21
Add dom safety to telephone

StevePole:


StevePole:
The issue is I have absolutely no idea why it wouldn't be there.  It's part of the template, there is no logic that says "only show if...", so it should not be possible for it to not exist.  BUT... in this one particular rollbar example, something happened for this user, and it was not there.  I can't write a test for it because it will always be there in the tests.  The code change just means if that scenario ever reoccurs, the page will not fail.

PR #187, opened 2018-05-21
Add icon component

OFFLlNE:
image


demo/index.html
```diff
@@ -28,7 +28,7 @@
     <div class="row m-t-3">
 
       <div class="col-lg-3 col-lg-push-9">
-        <nav data-spy="affix" data-offset-top="0">
+        <nav>
```
oliverviljamaa: Can you explain that change?

package.json
```diff
@@ -43,29 +43,32 @@
   },
   "homepage": "https://github.com/transferwise/styleguide-components",
   "dependencies": {
-    "angular": "^1.5.0",
-    "jquery": "^3.2.1"
+    "angular": "^1.7.0",
```
oliverviljamaa: I think we should be consistent with our Angular versions, and not all apps support later than 1.5.

package.json
```diff
@@ -43,29 +43,32 @@
...
+    "eslint-plugin-import": "^2.12.0",
     "gh-pages": "^1.1.0",
     "git-directory-deploy": "^1.5.1",
-    "html-loader": "^0.5.1",
+    "html-loader": "^0.5.5",
     "htmllint": "^0.6.0",
     "htmllint-cli": "0.0.6",
     "htmllint-loader": "^1.3.8",
-    "jasmine-core": "^2.8.0",
+    "install-peers": "^1.0.2",
```
oliverviljamaa: Do we need this?

package.json
```diff
@@ -74,17 +77,24 @@
...
-    "karma-mocha-reporter": "^2.2.4",
+    "karma-mocha-reporter": "^2.2.5",
     "karma-phantomjs-launcher": "^1.0.0",
     "karma-safari-launcher": "^1.0.0",
-    "less": "^2.7.2",
+    "less": "^2.7.3",
+    "less-loader": "^4.1.0",
     "live-server": "^1.1.0",
-    "phantomjs-prebuilt": "^2.1.5",
+    "mini-css-extract-plugin": "^0.4.0",
```
oliverviljamaa: What's the benefit of the mini?

package.json
```diff
@@ -74,17 +77,24 @@
...
-    "webpack": "^2.5.1",
-    "webpack-dev-server": "^2.9.1"
+    "uglify-js": "^3.3.26",
+    "uglifyjs-webpack-plugin": "^1.2.5",
+    "webpack": "^4.8.3",
+    "webpack-cli": "^2.1.3",
+    "webpack-dev-server": "^2.11.2"
+  },
+  "peerDependencies": {
+    "@transferwise/icons": "^0.11.2"
```
oliverviljamaa: Where do we add it as an external to ensure it not still ending up in the bundle?

src/layout/icon/icon.component.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  spriteIsOnPage() {
+    return !!this.document.querySelector('.svg-icon-sprite-container');
+  }
+
+  appendSpriteToPage() {
+    const div = this.document.createElement('div');
+    div.style.display = 'none';
+    div.classList.add('svg-icon-sprite-container');
```
oliverviljamaa: Can we not repeat the class name?

demo/index.html
OFFLlNE: Navbar was so full, that you couldn't see some parts of the nav.
Those were not accessible:
image

src/layout/icon/icon.demo.html
```diff
@@ -0,0 +1,40 @@
+<h4 id="icons">Icons</h4>
+
+<div class="row">
+  <div class="col-md-8">
+    <p>
+      <code>tw-icon</code> is a component for different icons. You can see all the available icons
+      <a href="https://transferwise.github.io/icons/" target="_blank">here</a>.
```
oliverviljamaa: Nice.

package.json
OFFLlNE: So we should remove the `hat` from the version? and keep it as 1.5.0 for ever/now

src/layout/icon/icon.demo.html
```diff
@@ -0,0 +1,40 @@
...
+    </p>
+
+    <tw-icon
```
oliverviljamaa: I think these would be more readable when they were all inline.

package.json
OFFLlNE: Nope, good call. While commiting forgot to remove

src/layout/icon/icon.demo.html
oliverviljamaa: Does it make sense to show them all without labels underneath? How to tell which one is which on the demo page? Maybe a select for size would suffice?

package.json
OFFLlNE: `Usually, it's recommended to extract the style sheets into a dedicated file in production using the MiniCssExtractPlugin. This way your styles are not dependent on JavaScript.` From less-loader readme

src/demo.js
```diff
@@ -86,7 +86,9 @@ module.config(['$compileProvider', ($compileProvider) => {
 
     <h5>Layout</h5>
     <ul class="list-unstyled">
+      <li><a href="index.html#affix">Affix</a></li>
       <li><a href="index.html#cards">Cards</a></li>
+      <li><a href="index.html#icons">Icons</a></li>
```
oliverviljamaa: Are icons layout?

package.json
OFFLlNE: ¯\_(ツ)_/¯

src/demo.js
OFFLlNE: Seemed the most suitable from all the existing topics

src/layout/icon/icon.less
```diff
@@ -0,0 +1,5 @@
+@import (inline) '../../../node_modules/@transferwise/icons/dist/icons.min.css';
```
oliverviljamaa: I don't think we should have the discrepancy between scripts and styles here – if we won't bundle the scripts and require it as a peer, it should be the same for styles.

webpack.config.js
```diff
@@ -1,99 +1,106 @@
 const path = require('path');
+
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 
+const devMode = process.env.NODE_ENV !== 'production';
```
oliverviljamaa: `isDevelopment`?

src/layout/icon/icon.component.js
OFFLlNE: Tried, had some errors when trying to use the const. Will try again

package.json
oliverviljamaa: A tilde (~) matches the newest minor version, so `"~1.5.0"` would be best, as Angular doesn't do breaking changes on patches, but does on minors.

package.json
oliverviljamaa: Interesting, we're just always using [extract-text-webpack-plugin
](https://www.npmjs.com/package/extract-text-webpack-plugin), as it's more popular. Why do we need the styles in a separate file, though?

package.json
oliverviljamaa: Let's not forget that!

src/demo.js
oliverviljamaa: A new one is needed then, I guess.
`Graphics`?

package.json
OFFLlNE: I feel like while trying to use the classical `extract-text-webpack-plugin` I ran into problems with webpack 4 and suggested solution was to use this plugin. I might give it another go, since other things are fixed in the config, maybe get lucky

PR #191, opened 2018-05-24
Fix issue with tw-select not updating selected value on model change

sikumiku:
Change directive to use one-way binding and fix bug in code to allow for selected value to update with external model change.

### DIFF
--- a/src/forms/select/select.controller.js
+++ b/src/forms/select/select.controller.js
@@ -10,7 +10,7 @@ class SelectController {
     this.dom = TwDomService;
 
     this.element = $element[0];
-    $scope.$watch('$ctrl.ngModel', (newValue, oldValue) => this.onModelChange(newValue, oldValue), true);
+    $scope.$watch('$ctrl.ngModel', (newValue, oldValue) => this.onModelChange(newValue, oldValue));
   }
 
   $onInit() {
--- a/src/forms/select/select.controller.js
+++ b/src/forms/select/select.controller.js
@@ -10,7 +10,10 @@ class SelectController {
     this.dom = TwDomService;
 
     this.element = $element[0];
-    $scope.$watch(this.$ngModel, this.onModelChange);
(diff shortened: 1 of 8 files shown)


src/forms/select/select.spec.js
```diff
@@ -76,6 +76,13 @@ describe('Select', function() {
         it('should set the value of hidden control', function() {
           expect(component.find(SELECT_SELECTOR).val()).toBe('2');
         });
+        describe('when ngModel changes', function() {
+          it('should show new model as selected', function () {
+            $scope.ngModel = '4';
+            $scope.$digest();
+            expect(component.find(SELECT_SELECTOR).val()).toBe('4');
+          });
```
StevePole: Would be cool to add another one like this:
```
it('should update the selected text' function() {
  var selectedTextElement = component.find(SELECTED_LABEL_SELECTOR)[0];
  expect(selectedTextElement.innerText.trim()).toBe('Four');	           
});
```

src/forms/select/select.controller.js
```diff
@@ -10,6 +10,21 @@ class SelectController {
     this.dom = TwDomService;
 
     this.element = $element[0];
+    $scope.$watch('$ctrl.ngModel', (newValue, oldValue) => {
```
StevePole: I think it was nice to have the onModelChange function separately, it will help keep the constructor from getting cluttered.

src/forms/select/select.controller.js
```diff
@@ -10,6 +10,7 @@ class SelectController {
     this.dom = TwDomService;
 
     this.element = $element[0];
+    $scope.$watch('$ctrl.ngModel', (newValue, oldValue) => this.onModelChange(newValue, oldValue), true);
```
StevePole: I still think we can do this `$scope.$watch('$ctrl.ngModel', this.onModelChange);` Did you try it?

PR #194, opened 2018-07-06
Fix tw-select behavior when filter has not result and return pressed

lapostoj:


src/forms/select/select.spec.js
```diff
@@ -580,6 +580,18 @@ describe('Select', function() {
...
+      $filterInput.val('a').trigger('change');
+      var selectedElements = component.find('.active');
+      expect(selectedElements.length).toBe(0);
+
+      triggerKeyCode(filterInput, SPECIAL_KEYS.return);
+
+      var selectedOptionElement = selectedElements[0];
+      var OptionOne = OPTIONS[1];
+      expect($scope.ngModel).toBe(OptionOne.value);
+      expect($(selectedOptionElement).text().trim()).toBe('');
```
StevePole: selectedOptionElement.innerText.trim()

src/forms/select/select.spec.js
```diff
@@ -580,6 +580,18 @@ describe('Select', function() {
...
     });
+    it('should prevent default if no matching option and return key pressed', function() {
+      $filterInput.val('a').trigger('change');
+      var selectedElements = component.find('.active');
+      expect(selectedElements.length).toBe(0);
+
+      triggerKeyCode(filterInput, SPECIAL_KEYS.return);
+
+      var selectedOptionElement = selectedElements[0];
+      var OptionOne = OPTIONS[1];
```
StevePole: 'optionOne'

PR #198, opened 2018-07-13
Update ci config

lapostoj:


.circleci/config.yml
```diff
@@ -4,23 +4,23 @@ jobs:
   test:
     working_directory: ~/styleguide-components
```
OFFLlNE: lets do
```
defaults: &defaults
  working_directory: ~/styleguide-components
  docker:
    - image: circleci/node:10.6-browsers
```

lapostoj:
htmllint-loader installation was failing on circleci and the repo seems to be archived anyway... https://github.com/robbiedigital/htmllint-loader

Overall why do we need any linting with webpack? Linting is not really a build time concern I'd say.

PR #199, opened 2018-07-16
Update circle config npm tokens

OFFLlNE:
- Replaced npm install to npm ci
- Upgrade circleci workflow


lapostoj:
oh I had this one already...
https://github.com/transferwise/styleguide-components/pull/198

But it's failing because the `circleci.com` has not been configured with the new token and contexts

OFFLlNE:
Oh lul, should've checked 😂 I will add commets to yours @lapostoj :D

PR #201, opened 2018-09-27
Add a new layout property to tw-definition list

vahurh:
- adds a new `layout` property, which defaults to 'vertical'
- adds two new configurable layouts, `justified` and `horizontal`


src/forms/definition-list/definition-list.component.js
```diff
@@ -9,7 +9,8 @@ const DefinitionList = {
     initialFields: '<fields',
     locale: '@',
     title: '@',
-    narrow: '<'
+    narrow: '<',
+    layout: '<'
```
StevePole: Let's make this an '@', then you won't need double strings in here:
```
<tw-definition-list
  layout="'horizontal'"
```
A couple rules of thumb I apply: 
- Will it change dynamically or will this use of the component always use the same value?  
  - If a more static thing it's weird to pass a variable.
- Is it always a string?
  - Then just easier to pass as a string. (like locale and title above)

src/forms/definition-list/definition-list.html
```diff
@@ -5,13 +5,14 @@
     class="page-header p-t-3">
   </h4>
 </div>
-<div class="row row-equal-height">
+<div class="row" ng-class="{'row-equal-height': $ctrl.layout === 'vertical'}">
```
StevePole: Can you explain the usage of `.row-equal-height` to me?  I wonder if it might be better to make the usage of `.row` and `.col-xs-12` conditional on us not being in justified mode, rather than compensating with the addition of this?

vahurh: `row-equal-height` is necessary in vertical mode to make the `col-sm-6`-s in one row be of the same height. 

The reason it's using  `divs` as opposed to having `col-sm-6` on the  `dl` as is in the bootstrap demos is that it supports multiple formats of fieldset structures and uses a different `dl` depending on the format. 

But correct - we can make row & col-xs-12 also conditional. Will add it.

PR #203, opened 2018-10-02
Call onRefreshRequirements in onChange for every field type

Kaarel94:
changes:
- onRefreshRequirements called in onChange for every field type now and field events debounced for text fields so we don't make a call for every keystroke (unless the user types slowly). This is done to avoid a situation where a user clicks on the form submit button when a field has focus which would cause the onRefreshRequirements api call (in onBlur) to be called simultaneously with submitting the form.
- removed onRefreshRequirements from onBlur since we are now calling it in onChange for every field anyway so it isn't needed in onBlur
- moved the onRefreshRequirements call to after the model changes so that the call would have to most up to date values. Previously the onRefreshRequirements call was made with old values.


src/forms/fieldset/fieldset.controller.js
```diff
@@ -83,18 +73,14 @@ class FieldsetController {
...
       }
+
+      if (field.refreshRequirementsOnChange && this.onRefreshRequirements) {
+        this.onRefreshRequirements();
+      }
     });
   }
 }
 
-function controlRefreshesOnChange(control) {
```
StevePole: Could you explain why this is no longer necessary?

Kaarel94: This was only used to check for which types of fields to trigger onRefreshRequirements in the fieldChange method. Now that this we trigger onRefreshRequirements in fieldChange for every type of field this check is no longer necessary.

src/forms/fieldset/fieldset.spec.js
```diff
@@ -52,19 +52,6 @@ describe('Fieldset', function() {
     });
   });
 
-  describe('when a field has refreshRequirementsOnChange: true', function() {
```
vahurh: Do we have any other tests for refreshRequirementsOnChange? If no, we probably should

src/forms/fieldset/fieldset.controller.js
StevePole: I got that, but why did we restrict it before and we no longer do.  I'm pretty sure I made that decision, I don't remember why, but I must have had a reason...

StevePole: Oh, I see, I previously prevented it on textual/numeric input, but now it's ok to enable.  Fine 👍

src/forms/fieldset/fieldset.spec.js
StevePole: There isn't a lot to test, all it does is trigger a handler.  However we should tests that, so not sure we can remove this...

StevePole:
What's the status on this?  Can we put that test back, fix the conflict and merge?

PR #206, opened 2018-10-19
Add new popover placements to the pop over service

mstaicu:
Add new logic to the pop over service that will accommodate the new positioning system, which is composed of 4 main positions (top, right, bottom, left) and 3 sub positions (either, left, middle, right or top, middle, bottom). For more details check the Bootstrap [#184](https://github.com/transferwise/bootstrap/pull/184) pull request.

### DIFF
--- a/src/help/pop-over/pop-over.service.js
+++ b/src/help/pop-over/pop-over.service.js
@@ -1,804 +1,563 @@
-function PopoverService($rootScope) {
-  /**
-   * Register the global event listeners for clicks on the body element and
-   * viewport resizing
-   */
-  registerGlobalEventListeners();
+class PopOverService {
+  constructor($rootScope) {
+    this.popover = null;
+    this.$rootScope = $rootScope;
 
-  const BODY = document.getElementsByTagName('body')[0];
+    this.documentCallback = this.documentCallback.bind(this);
+    this.resizeCallback = this.resizeCallback.bind(this);
 
-  let popover = null;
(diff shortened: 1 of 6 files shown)


src/help/pop-over/pop-over.service.js
```diff
@@ -252,7 +279,36 @@ function PopoverService($rootScope) {
       };
     }
 
-    if (placement === 'right' || placement === 'right-top') {
+    if (placement === 'top-right') {
```
StevePole: This is now a huge method, can you refactor it into smaller chunks?

src/help/pop-over/pop-over.service.spec.js
```diff
@@ -1,99 +1,245 @@
...
+
+      popover = PopoverService.showPopover(config);
     });
 
-    it('should append the popover and make it visible', function () {
-      var popoverAppended = $window.document.body.contains(popover);
-      var popoverVisible = !popover.classList.contains('scale-down');
-      var popoverVisibility = popoverAppended && popoverVisible;
+    it('should append the popover and make it visible', function() {
+      const popoverAppended = document.body.contains(popover);
```
StevePole: Have you run the tests locally?  I'm pretty sure using const fails on my machine as the tests are not run through babel.

mstaicu: They seem to pass, on my local machine and on Circle CI

PR #208, opened 2018-11-05
remove redundant upload switch case from form-control

Kaarel94:
Remove the upload switch case from form-control since it is redundant.
Looking at the getControlType logic in https://github.com/transferwise/styleguide-components/blob/master/src/services/requirements/requirements.service.js#L379 that is used by the field component, 'upload' is never returned and so the field component will never pass down 'upload' type to form-control, making it redundant.

### DIFF
--- a/src/forms/form-control/form-control.spec.js
+++ b/src/forms/form-control/form-control.spec.js
@@ -205,6 +205,28 @@ describe('FormControl', function() {
     }, 1);
   });
 
+  describe('type: file - validation', function() {
+    beforeEach(function() {
+      $scope.model = null;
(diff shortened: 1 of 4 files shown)


src/forms/form-control/form-control.spec.js
```diff
@@ -205,28 +205,6 @@ describe('FormControl', function() {
     }, 1);
   });
 
-  describe('type: upload - validation', function() {
```
StevePole: Not sure if we should remove this test, presumably this still passes because of the legacy support in the requirements service?  Do we have another test that renders tw-upload?  Using type=file or base64url?

PR #209, opened 2018-11-13
Disable autocomplete when helpText present

StevePole:


src/forms/form-control/form-control.html
```diff
@@ -122,7 +122,8 @@
     ng-focus="$ctrl.focus()"
     ng-blur="$ctrl.blur()"
     ng-min="$ctrl.ngMin"
-    ng-max="$ctrl.ngMax" />
+    ng-max="$ctrl.ngMax"
+    autocomplete="{{ $ctrl.helpText ? 'disabled' : 'on' }}" />
```
mikknarusk: I've never seen `autocomplete="disabled"`, shouldn't it be `off`?

StevePole: Good question!  Unfortunately, Chrome ignores autocomplete off, the only way to disable it is to specify an 'invalid' value, which 'disabled' is, I just chose it because it as a readable option that's meaning was clear.  I would have left a comment, but prefer not to in markup.  I'll move this to a controller function and comment there.

PR #218, opened 2019-02-21
Full support for JSON schema dynamic forms

StevePole:
Adds support for JSON schema forms.  This functionality is not included in the main bundle, but in a new library file.

This follows the implementation guide here, which will help you understand what's going on: https://transferwise.atlassian.net/wiki/spaces/DF/pages/921600151/Implementation+Guide

Effectively we build components to handle the different types of schemas we might encounter:

- Strings
- Numbers
- Integers
- Booleans

- Objects
- Arrays
- allOfs
- oneOfs

We also build a generic schema component that can handle all of these, and delegate responsibility to the specific components (you can think of this component like a switch statement).

The complex schema types (object, array, allOf, oneOf) contain nested schemas (e.g. what types the object properties have, or what is the structure of the items in the array), so we use the generic schema component recursively to ensure we can render these nested schemas, no matter what type they are.

Each component renders a form corresponding to the JSON schema that is receives.  By filling in the form, a matching JSON structure is constructed.  The new JSON structure is broadcast out using an onChange handler.  It is up to the consumer of these component to handle that JSON, presumably posting it to an API, but this is not within scope of these changes.

complex

recipient


src/json-schema/array-schema/controller.js
```diff
@@ -0,0 +1,44 @@
+const simpleTypes = ['string', 'number', 'integer', 'boolean'];
+
+class Controller {
+  $onInit() {
+    if (!Array.isArray(this.model)) {
+      if (this.schema && this.schema.items && this.isSimpleType(this.schema.items.type)) {
+        this.model = [];
+      } else {
+        this.model = [{}];
```
skywritergr: Is this basically saying that if it's not a simple type then it has to be an object?

src/json-schema/array-schema/controller.js
```diff
@@ -0,0 +1,44 @@
...
+      }
+    }
+  }
+
+  onModelChange(model, index) {
+    this.model[index] = model;
+    if (this.onChange) {
```
skywritergr: As I mentioned above this pattern of looking for that function and then use it is confusing. You do get eventually that it's coming from the bindings but it takes some time. Same applies across the PR. Maybe it's just a case of better naming.

src/json-schema/array-schema/controller.js
```diff
@@ -0,0 +1,44 @@
...
+      this.onChange({ model: this.model });
+    }
+  }
+
+
+  onRefreshRequirements(model, index) {
```
skywritergr: Also again, the two functions are almost the same function.

src/json-schema/array-schema/template.html
```diff
@@ -0,0 +1,40 @@
+<div ng-if="$ctrl.isSimpleType($ctrl.schema.items.type)">
+  Unsupported
```
skywritergr: Is this supposed to appear in production?

src/json-schema/basic-type-schema/controller.js
```diff
@@ -0,0 +1,30 @@
...
+    if (this.onChange) {
+      this.onChange({ model });
+    }
+
+    if (this.schema.refreshRequirementsOnChange && this.onRefresh) {
+      this.onRefresh({ model });
+    }
+  }
+
+  getValidationMessages() {
```
skywritergr: So this can return either a string, or an array or a boolean? I would prefer if it was just one and then the consumer of this function deal with this.

src/json-schema/generic-schema/template.html
```diff
@@ -0,0 +1,56 @@
+<one-of-schema
```
skywritergr: Would be lovely if could use a switch here. https://code.angularjs.org/1.6.10/docs/api/ng/directive/ngSwitch

src/json-schema/object-schema/template.html
```diff
@@ -0,0 +1,34 @@
...
+  </p>
+  <div ng-if="$ctrl.schema.help.warning"
+    class="alert alert-warning">
+    {{ $ctrl.schema.help.warning }}
+  </div>
+
+  <div class="row">
+    <div ng-repeat="(property, nestedSchema) in $ctrl.schema.properties"
+      class="col-xs-12"
+      ng-hide="nestedSchema.hidden"
```
skywritergr: Why ng-hide and not ng-if?

src/json-schema/utils/clean-model.js
```diff
@@ -0,0 +1,93 @@
...
+    return model;
+  }
+  return; // eslint-disable-line
+}
+
+function cleanModelWithBooleanSchema(model) {
+  if (typeof model === 'boolean') {
+    return model;
+  }
+  return; // eslint-disable-line
```
skywritergr: In all of those returns I would return null and handle it. (or don't handle, but at least return null)

src/json-schema/utils/clean-model.js
```diff
@@ -0,0 +1,93 @@
+
+function cleanModel(model, schema) {
```
skywritergr: The terminology "clean" is not clear to me what it's suppose to do or clean. Can we clarify further either with a different name or a comment?

src/json-schema/all-of-schema/controller.js
```diff
@@ -0,0 +1,50 @@
...
+  $onChanges(changes) {
+    if (changes.schema || changes.model) {
+      // We keep each model separately, and combine them for broadcast, this gives
+      // better control over cleaning up the model as we deal with values that are
+      // removed.
+      this.models = splitModel(this.model, this.schema ? this.schema.allOf : []);
+    }
+  }
+
+  onModelChange(model, index) {
```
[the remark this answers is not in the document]
StevePole: It can't be pure.  This component has several children, each in control of parts of the overall model. We need to maintain the state for each child, so that we can broadcast the combined model when just one of the children changes.

src/json-schema/array-schema/template.html
StevePole: Right now, we don't have any cases where we support arrays of simple types, this is just there to make that clear to a consumer.  In future I do expect us to add some support.

src/json-schema/basic-type-schema/controller.js
StevePole: Only an object or false.  But perhaps false is the wrong choice...

src/json-schema/generic-schema/template.html
StevePole: Unfortunately we can't as it is not dependent on one value.  We need to check for `allOf`, `oneOf`, and `type`

src/json-schema/object-schema/template.html
StevePole: In the case of hidden input fields we still render a hidden input to maintain support for regular HTML form posts.  However, if we do that, we don't want to render a field label, so we also need to hide the surrounding block.

src/json-schema/all-of-schema/controller.js
```diff
@@ -0,0 +1,50 @@
...
+    const schemaForIndex = this.schema.allOf[index];
+
+    this.models[index] = cleanModel(model, schemaForIndex);
+
+    if (this.onChange) {
+      this.onChange({ model: combineModels(this.models) });
+    }
+  }
+
+  onRefreshRequirements(model, index) {
```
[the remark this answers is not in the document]
StevePole: It's not straightforward as they trigger call chains up through the component stack, but the call chains start in different circumstances (e.g. only when refreshRequirementsOnChange is true).  However, I do have one idea.  I could change all of the onChange handlers to pass back, either, the new model and a 'refresh' boolean.  Or, the new model and the schema that triggered the change.  Then the consumer could interrogate the schema for refreshRequirementsOnChange.  I could then lose the onRefresh callbacks completely.  It's a fraction harder for the consumer, but cleaner in implementation.

src/json-schema/basic-type-schema/controller.js
StevePole: Changed to null

PR #219, opened 2019-02-25
fix(upload): Unable to remove file if cancelText is not provided [USD-3124]

joebartels-tw:
When `cancelText` is not provided to the upload component: the user has no way to change the file they uploaded.

**Current UI, with `cancelText` _(this behavior will remain the same)_:**
with cancelText

**Updated UI, with missing `cancelText`:**
missing cancelText

### DIFF
--- a/src/forms/upload/upload.html
+++ b/src/forms/upload/upload.html
@@ -71,8 +71,21 @@
         <span class="icon icon-alert icon-xxl text-danger m-b-1"></span>
       </div>
       <div ng-if="$ctrl.hasTranscluded" ng-transclude></div>
-      <p ng-if="$ctrl.cancelText" class="m-t-2 m-b-0">
-        <a href="" ng-click="$ctrl.clear()">{{$ctrl.cancelText}}</a>
+      <p class="m-t-2 m-b-0">
+        <a href="" ng-click="$ctrl.clear()">
+          <svg 
+            ng-if="!$ctrl.cancelText" 
+            viewBox="0 0 24 24" 
+            fill="#00b9ff"
+            xmlns="http://www.w3.org/2000/svg"
+            height="25">
+            <path d="M12 23a11 11 0 1 1 11-11 11 11 0 0 1-11 11zm0-21a10 10 0 1 0 10 10A10 10 0 0 0 12 2z"></path>
+            <path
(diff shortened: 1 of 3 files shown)


src/forms/upload/upload.html
```diff
@@ -71,8 +71,15 @@ <h4 class="m-b-2" ng-if="!$ctrl.isTooLarge && $ctrl.errorMessage">{{$ctrl.errorM
         <span class="icon icon-alert icon-xxl text-danger m-b-1"></span>
       </div>
       <div ng-if="$ctrl.hasTranscluded" ng-transclude></div>
-      <p ng-if="$ctrl.cancelText" class="m-t-2 m-b-0">
-        <a href="" ng-click="$ctrl.clear()">{{$ctrl.cancelText}}</a>
+      <p class="m-t-2 m-b-0">
+        <a href="" ng-click="$ctrl.clear()">
+          <span 
+            ng-if="!$ctrl.cancelText" 
+            class="glyphicon glyphicon-remove-circle"
```
StevePole: Is there anything in the `icon` set we can use?  Ideally we wouldn't use glyphicons, once we switch the caret over to something else, it would be good to remove them overall and save the transport cost.

StevePole: Looks we have something in here, but it's not a dependency.
https://transferwise.github.io/icons/

And here (this is a dependency, but only for the search icon in select):
https://transferwise.github.io/iconfont/

I don't really want to make importing an additional icon font a requirement, just to get one glyph.  I wonder if we can inline the SVG from the former.  I would also potentially do it for the search icon, and remove the icon dependency altogether.

(The select can also use icons within options, but these are non-essential and will fail silently if the icon font is not available)

StevePole: Or if you can use the SVG from the former repo, without bundling the whole thing?

joebartels-tw: Sure I can extract the **svg** from transferwise/icons

Another thought is to use placeholder text like "Remove this file" but I didn't see a precedence for using translations in the dynamic-flow... 

I'll move towards extracting the **svg** though, but open to thoughts on using a placeholder text.

joebartels-tw: @StevePole I updated the code to use `<svg>` from the transferwise/icons repo

I also updated the screenshot for a missing `cancelText` since it's a tad different icon.

PR #221, opened 2019-03-15
Change twFieldset to support new required format for twField

StevePole:


src/forms/fieldset/fieldset.controller.js
```diff
@@ -11,6 +11,9 @@ class FieldsetController {
     if (!this.model) {
       this.model = {};
     }
+    if (!this.requiredFields) {
+      this.requiredFields = [];
```
joebartels-tw: how do you feel about
```js
if (!Array.isArray(this.requiredFields)) { 
  this.requiredFields = [];
}
```

only reason I suggest is b/c there are a couple places where `this.requiredFields` is being subtly type checked...

line 46 could turn into `if (!this.requiredFields.length)`
line 97 could turn into `return this.requiredFields.indexOf(key) >= 0;`

granted, I'm not well versed on lifecycle of angular components so not sure if either of those checks could happen before the `$onInit`

StevePole: It's a good question.  $onInit fires once the binding are established, and before $onChanges, so it's safe.  It's basically like putting something in the constructor, but with the guarantee that the bindings will be ready.

src/services/requirements/requirements.service.js
```diff
@@ -406,9 +406,21 @@ function RequirementsService($http) {
...
+/**
+ * Older style format had required as a property of fields, in JSON schema it
+ * should be a separate array of property names
+ */
+function getRequiredFields(fields) {
+  if (Array.isArray(fields)) {
+    throw new Error('Expected field map');
+  }
+  // Return array of keys that have required set
+  return Object.keys(fields).filter(property => fields[property].required);
```
joebartels-tw: won't this just be an array of array indices? 

e.g. I thought the older style looked like this:
```
fields: [
  { key: "basicText", "required": true },
  { key: "notRequired", "required": false }
]
```
But I see the spec is using, what I thought, was the newer json-schema inspired structure?

joebartels-tw: ooh nvm. I see this is not meant for the old style... I misread the `if (Array.isArray(fields))` part... ignore me

StevePole: :-)  Yes, originally it was an array, the RequirementsService makes sure it's in JSON schema format before we start doing anything with it.

PR #223, opened 2019-04-09
Improve JSON schema form documentation

StevePole:
We also add a new component `json-schema-form`, this is a synonym for `generic-schema-component` so uses the same implementation, but presents a better name for consumers and gives us the potential to let the internal functionality diverge from the external in future.  One case I have in mind, is that it might be useful to know how deep we are in the schema.  However I don't want consumers to need to worry about that, by passing in a '1' (or 0).  

Also addresses a small bug with browser autofill.

image

image


src/json-schema/demo.js
```diff
@@ -1,38 +1,25 @@
...
+  }
+
   $onInit() {
+    this.locale = 'en-GB';
+
     this.errors = {
-      currency: 'This one is bad',
-      accountNumber: 'Also this one',
-      streetAddress: 'And this one could be better',
+      industry: 'We were hping you would pick the other one',
```
skywritergr: Typo :)

PR #227, opened 2019-05-29
Adjust getDateString to get exact time if in 6days

OFFLlNE:
Before if the date was more than 48hours away, then we formatted out the exact time and just displayed the date for medium/short.
With this change we will show the time if the date is within 6days (144 hours).

Was:
image

Will:
image

Personally do not see any downsides to that, only being more transparent with time

### DIFF
--- a/src/services/date/date.service.js
+++ b/src/services/date/date.service.js
@@ -349,10 +349,10 @@ function DateService() {
     currentYear,
     format
   ) => {
-    const sixDaysInHours = 144 * 60 * 60 * 1000;
+    const sixDays = 144 * 60 * 60 * 1000;
     const eightDays = 8 * 24 * 60 * 60 * 1000;
 
-    const hasTime = Math.abs(offset) < sixDaysInHours && format !== 'long';
+    const hasTime = Math.abs(offset) < sixDays && format !== 'long';
     const hasDate = !hasTime;
     const hasWeekday = Math.abs(offset) < eightDays || format === 'long';
     const hasMonth = !hasWeekday || format === 'long';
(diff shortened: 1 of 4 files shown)


src/services/date/date.service.js
```diff
@@ -349,10 +349,10 @@ function DateService() {
     currentYear,
     format
   ) => {
-    const fourtyEightHours = 48 * 60 * 60 * 1000;
+    const sixDaysInHours = 144 * 60 * 60 * 1000;
```
rs6g10: For better readability the protocol is already implemented in below line:
```
sixDaysInHours = 6 * 24 * 60 * 60 * 1000
```

rs6g10: Yea; `sixDays` 😃

StevePole:
What motivated this?  We rarely know when things will happen with such precision of they are further away, that's why the time was omitted.

There is a proposal to change this formatter to allow an optional 'max-precision' parameter, here: https://transferwise.atlassian.net/wiki/spaces/TE/pages/1017054271/Date+formatting

OFFLlNE:
@StevePole Oh nice, I did not know this has such beautiful documentation on it. Do you want me to update it as well?

But the reason why we did the change was that quite often guaranteed rate is promised to a date which is 48+ hours away. Which causes problems we display "We need to receive the money on Monday 3rd" So it gives them the idea that they have the whole day, but actually it would be something like `Monday 3rd 1PM` or something similar. And since this was our error we had to refund if people started complaining about that (We did send on 3rd as your website asked)

PR #228, opened 2019-05-31
ASIA-2103: Improve tw-select performance when it has a large number of options

sneakertack:
We were informed that the UI would freeze when someone was trying to add a Japanese recipient for a to-JPY transfer: https://transferwise.slack.com/archives/C0JNUL7RD/p1558104250003800. It turns out Japan would generate the largest amount of entries for bank or branch codes (over 3000). There are multiple things that can be improved but one of them was that `tw-select` would create 3000 elements on the page.

This PR improves `tw-select` for when there's such a high number of options. Specifically, it will only render the first 50 entries, with a `...` at the bottom. Click it and the list expands to 100. Click it again and the list expands to 150 etc. The `...` doesn't appear if there's less than 51 options.

### DIFF
--- a/src/forms/select/select.html
+++ b/src/forms/select/select.html
@@ -113,7 +113,7 @@
         'dropdown-header': option.header,
         'tw-select-option': !option.header && !option.disabled
       }">
-      <span ng-if="option.header" class="text-ellipsis">{{option.header}}</span>
+      <span ng-if="option.header" class="text-ellipsis" ng-click="$event.stopPropagation()">{{option.header}}</span>
       <a href=""
         ng-if="!option.header"
         ng-click="$ctrl.optionClick(option, $event)"
@@ -148,6 +148,16 @@
           ng-if="option.note" class="tw-select-note small m-l-1">{{option.note}}</span><span
           ng-if="option.secondary"
           class="tw-select-secondary small text-ellipsis">{{option.secondary}}</span>
+      </a>
+    </li>
+
+    <li ng-if="$ctrl.hasMoreOptionsToReveal">
+      <a href=""
+        ng-click="$ctrl.revealMoreOptions($event)"
+        class="tw-select-load-more"
+        tabindex="-1"
+        tw-focusable >
+        ...
       </a>
     </li>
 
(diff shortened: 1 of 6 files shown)


src/forms/select/select.html
```diff
@@ -113,7 +113,7 @@
         'dropdown-header': option.header,
         'tw-select-option': !option.header && !option.disabled
       }">
-      <span ng-if="option.header" class="text-ellipsis">{{option.header}}</span>
+      <span ng-if="option.header" class="text-ellipsis" ng-click="$event.stopPropagation()">{{option.header}}</span>
```
sneakertack: Minor driveby fix: Previously, clicking on a header within the dropdown would close the list, due to causing a blur event. Now, it won't close the list (the user should have to click out of the list in order to close it).

src/forms/select/select.spec.js
```diff
@@ -508,22 +509,50 @@ describe('Select', function() {
...
+
+    it('should limit the number of options initially shown', function () {
+      expect(component.find(LIST_ITEMS_SELECTOR).length).toBe(50);
+    });
+
+    it('should show a "..." option at the end to load more options', function () {
+      expect(component.find(OPTION_LOAD_MORE_SELECTOR).length).toBe(1);
+    });
+
+    it('should load more options when "..." is clicked', function () {
```
StevePole: Nice tests 👏   If I was being super picky :-) I'd put the 'when' bit in a describe:
```
describe('when the "..." is clicked', function() {
  beforeEach(function() {
    component.find(OPTION_LOAD_MORE_SELECTOR).click();
  });
  it('should display more options', function() {
    expect(component.find(LIST_ITEMS_SELECTOR).length).toBe(100);
    expect(component.find(OPTION_LOAD_MORE_SELECTOR).length).toBe(1);
  });
  describe('when the "..." is clicked to show the remaining options', function() {
    beforeEach(function() {
      component.find(OPTION_LOAD_MORE_SELECTOR).click();
    });
    it('should display the remaining options', function() { ... });
    it('should hide the "..." link', function() { ... });
  });
});
```

sneakertack: The clicking twice in the current test is very convenient for testing both has-more-options then no-more-options behaviour :grimacing:. if we nest it in 2 describes to achieve the 2nd click's behaviour, it could be a bit hard for other readers to understand what's going on at a glance. So will leave it as it is in this case.

PR #229, opened 2019-06-06
[ASIA-1973] Live camera upload component

HaoYuanTW:
This PR adds a live camera upload component for JPY verification requirement. 

Currently we have an uploading component relying on <input> tag, we cannot reuse the existing upload component for 2 reasons

1. Verification requirement states user HAVE to upload photos with a live camera, picking from album is forbidden. Input tag has parameter that supports this, but the parameter is not strictly respected, we will risk breaking compliance. 

2. JPY verification has a requirement to upload a "tilted" ID with thickness of ID card and some portion of front side of ID card visible. It is hard for users to figure out how to do this. Based on our user testing, we found it is crucial to add a camera overlay guiding user to position their id cards properly. We didnt find any way to make this happen for the input tag based upload component. 

Screenshot 2019-08-09 at 14 53 54

---

At the moment we have built the component and tested it on android and IOS, there are some small problems related to how mobile browser communicate screen size, but overall looking alright. 

Gif showing the flow (cancel and upload)
ezgif com-video-to-gif (1)


StevePole:
Could you post screenshots/gif in the PR description?

HaoYuanTW:
@StevePole Yes, getting to it tmr

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,385 @@
...
+      })
+      .catch(() => {
+        isFullScreen = false;
+        console.error('Failed to acquire full screen.');
+      })
+      .finally(() => {
+        // After trying to acquire full screen, resolve video stream
+        function resolveScreenDimensions($ctrl) {
+          // TODO haoyuan : firefox is recognizing pixel's btm bar
+          console.log(`screen : ${window.screen.height} x ${window.screen.width}`);
```
sneakertack: There's a few debugging `console.log`s scattered through this PR, which we'll eventually have to remove before merge.

demo/index.html
```diff
@@ -30,9 +30,8 @@
       <div class="col-lg-3 col-lg-push-9">
         <nav data-spy="affix" data-offset-top="0">
           <docs-navigation></docs-navigation>
-
-          <h5>Data Model</h5>
-          <pre>{{vm.model | json}}</pre>
+          <!-- <h5>Data Model</h5> -->
+          <!-- <pre>{{vm.model | json}}</pre> -->
```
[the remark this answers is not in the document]
HaoYuanTW: This part is breaking testing on mobile, everytime something gets clicked, a huge section will be generated bumping everything else out of place. I intend to disable it for real.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
HaoYuanTW: If it is possible, I wish to merge this commit with the console logs for internal testing. 
2 tricky parts need observations here
1. It is not 100% clear how different phones support full screen mode
2. It is not clear if our screen dimension resolution works well under different conditions (i will pull a list of screensizes and make test cases for them soon)

sneakertack: It sounds helpful, but note that ideally no customer should open up his console and stumble upon such messages. Thus some options:

1) Either the logs only appear when a `<tw-upload-live-cam-docs>` is rendered (which doesn't happen for normal customers), or
2) there's some kind of debug flag set during testing, that's off by default

Instead of using direct `console` there's also the AngularJS `$log` wrapper, which is supposed to be the recommended way to go about it, maybe it might provide some facility to toggle debugging or somesuch.

src/forms/upload/camera-only-upload/camera-only-upload.less
```diff
@@ -0,0 +1,140 @@
+/**
+ Stolen from bootstrap directly, can remove this file if we upgrade the bootstrap component to 4
```
StevePole: I would like to migrate this into our Bootstrap so we do not need to add style to the build of this repo.  Existing clients will not be including any new styles.

src/forms/upload/camera-only-upload/camera-only-upload.component.js
```diff
@@ -0,0 +1,26 @@
...
+    // Angular binding
+    ngDisabled: '<', // TODO prolly not needed
+    ngModel: '=', // TODO prolly not needed
+    ngChange: '&', // TODO prolly not needed
+
+    name: '@', // TODO prolly not needed form name
+
+    errorMessage: '@', // TODO prolly not needed
+
+    showCaptureScreen: '=',
```
StevePole: Does this need two-way binding?  I would guess one-way is fine '<'

src/forms/upload/camera-only-upload/camera-only-upload.component.js
```diff
@@ -0,0 +1,26 @@
...
+    cameraFaceMode: '@', // environment/user
+
+    // Angular binding
+    ngDisabled: '<', // TODO prolly not needed
+    ngModel: '=', // TODO prolly not needed
+    ngChange: '&', // TODO prolly not needed
+
+    name: '@', // TODO prolly not needed form name
+
+    errorMessage: '@', // TODO prolly not needed
```
StevePole: I think you can delete all of these

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    this.$scope.$watch('$ctrl.showCaptureScreen', (newValue, oldValue) => {
+      if (newValue !== oldValue && newValue) {
+        this.onLiveCamFlowStart();
+      }
+    });
+
+    window.addEventListener('orientationchange', createOrientationChangeCallback(this), false);
+    const video = this.$element[0].querySelector('#video-preview #video');
```
StevePole: You do this in more than one place.  Can you set up a reference $onInit?

StevePole: In actual fact, you should consider moving some of this to an $onInit function.  OnInit runs after all bindings are ready, whereas the constructor may run before: https://ultimatecourses.com/blog/on-init-require-object-syntax-angular-component

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+      const $ngModel = this.$element.controller('ngModel');
+      if (!$ngModel.$setViewValue) {
+        return;
+      }
+      $ngModel.$setViewValue(value);
+    }
+  }
+}
+
+// TODO haoyuan : move this inside class?
```
StevePole: Either move inside the class or decouple from controller.  I would prefer to return an object something like:
```
{
  width: ...
  height: ...
  ...
}
```
You can use destructuring to make the assignment to the controller neat.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+function createVideoPlayCallback($ctrl) {
+  return function videoPlayCallback() {
+    function resolveVideoDimensions() {
```
StevePole: Confusing syntax.  Can you separate these functions?

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+  return !!(navigator.mediaDevices && navigator.mediaDevices.getUserMedia);
+}
+
+/**
+ * Resolve canvas x, y offset from screen edge and canvas size
+ * Function is based off the fact that video stream will expand
+ * to fill one dimension of the screen and be centered wrt the other dimension
+ * We need to know which dimension it is
+ */
+function resolveCanvasDimensions($ctrl, screenHeight, screenWidth, videoResHeight, videoResWidth) {
```
StevePole: Generally, I wonder if you can move out some generic canvas handling code to a separate service?  A lot of the named function don't seem coupled to the component controller.

src/forms/upload/upload.component.js
```diff
@@ -31,7 +31,12 @@ const Upload = {
     onSuccess: '=',
     onFailure: '=',
     onCancel: '=',
-    maxSize: '<' // TODO move to ngMax?
+    maxSize: '<', // TODO move to ngMax?
+
+    // Camera only upload options
+    source: '<', // Source of file to upload (file system? live camera only?)
```
StevePole: Is this an enum?  What are the potential values?

src/forms/upload/upload.demo.html
```diff
@@ -99,47 +103,75 @@ <h4 id="upload">Upload</h4>
...
           ng-click="$ctrl.makeFancy()">Make fancy</button>
       </h5>
+
+      <div class="form-group" ng-init="$ctrl.size = 'md'">
+        <label class="control-label">
+          File Source
+        </label>
+        <tw-select
+          placeholder="File Source"
+          ng-required="true"
```
StevePole: You don't need ng-required here.

src/forms/upload/upload.demo.html
```diff
@@ -99,47 +103,75 @@ <h4 id="upload">Upload</h4>
...
+          options="$ctrl.sources"></tw-select>
+      </div>
+
+      <div class="form-group" ng-init="$ctrl.size = 'md'">
+        <label class="control-label">
+          Camera Face Mode
+        </label>
+        <tw-select
+          placeholder="Camera Face Mode"
+          ng-required="true"
```
StevePole: You don't need ng-required here.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+  console.log(`canvas x-offset : ${canvasDimensions.xOffset}`);
+  console.log(`canvas y-offset : ${canvasDimensions.yOffset}`);
+  console.log(`canvas width to paint : ${canvasDimensions.paintWidth}`);
+  console.log(`canvas height to paint : ${canvasDimensions.paintHeight}`);
+
+  return canvasDimensions;
+}
+
+// Return full screen canvas for narrow screen layout
+function resolveNarrowCanvas($ctrl, screenHeight, screenWidth, videoResHeight, videoResWidth) {
```
StevePole: Don't pass $ctrl here.  Looks like you just need the videoHeight.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+    width: screenWidth,
+    xOffset: 0,
+    yOffset: 0,
+    // Do not draw overflow
+    paintHeight: parseInt(videoResHeight / ($ctrl.videoHeight / 100), 10),
+    paintWidth: videoResWidth
+  };
+}
+
+function resolveNormalCanvas(screenHeight, screenWidth, videoResHeight, videoResWidth) {
```
StevePole: Unclear naming, perhaps `getCanvasSizeAndOffset`?

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
StevePole: If you remove $ctrl from the other function (`resolveNarrowCanvas`), I don't think you need it here either.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
StevePole: The name of this function could be better.  I don't really know what it's for?  'resolve' sounds like a synonym for the canonical `get`.  Is this effectively `getCanvasSizeForNarrowScreens`?

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
StevePole: Looking more, it looks like we have two methods:
`getCanvasSizeAndPositionForPortraitScreen`
`getCanvasSizeAndPositionForLandscapeScreen`
?

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+  console.log('----- Recomputing canvas -----');
+  console.log(`screen width : ${screenWidth}`);
+  console.log(`screen height : ${screenHeight}`);
+  console.log(`video res width : ${videoResWidth}`);
+  console.log(`video res height : ${videoResHeight}`);
+
+  let canvasDimensions = null;
+
+  // handle layout where a landscape screen has a long and narrower camera window in it
+  if (screenHeight > screenWidth && videoResHeight / videoResWidth > screenHeight / screenWidth) {
```
StevePole: Rather than comment, encapsulate these in named functions that describe what they do.  Perhaps add a comment more about _why_ this is necessary, rather than when (which is described by the code).
```
if (screenIsPortrait(screenWidth, screenHeight) && 
  videoIsTallerThanScreen(screenWidth, screenHeight, videoWidth, videoHeight)) {
  ...
}
```

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+    this.showVideoPreview = false;
+    this.closeVideoStream();
+    const video = this.$element[0].querySelector('#video-preview #video');
+    video.srcObject = null;
+  }
+
+  onCaptureBtnClick() {
+    const canvas = this.$element[0].querySelector('#video-preview #display-canvas');
+    const video = this.$element[0].querySelector('#video-preview #video');
+    const resolved = resolveCanvasDimensions(
```
StevePole: Would be more readable with destructuring:
```
const { width, height, offsetX, offsetY } = getCanvasSizeAndPosition(...)
```
Otherwise I'm left wondering what that function returned and can only work out by reading more code.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+        isFullScreen = true;
+        console.log('Acquired full screen.');
+      })
+      .catch(() => {
+        isFullScreen = false;
+        console.error('Failed to acquire full screen.');
+      })
+      .finally(() => {
+        // After trying to acquire full screen, resolve video stream
+        function resolveScreenDimensions($ctrl) {
```
StevePole: Why do you need to create this function here?

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+          console.log(`screen inner : ${window.innerHeight} x ${window.innerWidth}`);
+
+          $ctrl.screenHeight = window.innerHeight;
+          $ctrl.screenWidth = window.innerWidth;
+        }
+
+        resolveScreenDimensions(this, isFullScreen);
+        console.log(`**screen resolved** : ${this.screenHeight} x ${this.screenWidth}`);
+
+        function postVideoStreamAcquisition($ctrl, stream) {
```
StevePole: What are we posting?  If this a handler for an event, use the canonical `on`, so `onVideoStreamAcquisition`

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+          $ctrl.mediaStream = stream;
+
+          /*
+           This is done instead of just reassigning video stream everytime
+           to prevent screen from blinking excessively during switch
+            */
+          if (video.srcObject !== $ctrl.mediaStream) {
```
StevePole: Where does `video` come from?  The scope inheritance is difficult to follow, is there a reason you can't pass in the parameters you need?

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,387 @@
...
+  }
+
+  onRecaptureBtnClick() {
+    this.onLiveCamFlowStart();
+  }
+
+  onUploadBtnClick() {
+    const displayCanvas = this.$element[0].querySelector('#video-preview #display-canvas');
+    const uploadCanvas = this.$element[0].querySelector('#video-preview #upload-canvas');
+    const resolved = resolveCanvasDimensions(
```
StevePole: You could also use destructuring to aid readability here.

demo/index.html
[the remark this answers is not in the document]
StevePole: I think you could just change it so that the upload component doesn't bind to this model.

src/forms/upload/upload.component.js
HaoYuanTW: yes, it is an enum value defined in dynamic interface, values are
CAMERA_ONLY (allow only camera upload)
FILE_ONLY (allow only upload from file system)
ANY (we dont care)
currently only CAMERA_ONLY is useful in DF.

src/forms/upload/camera-only-upload/camera-only-upload.less
HaoYuanTW: I remember we consulted frontend-guild at the time and were adviced to put CSS here because we are moving away from our bootstrap repo...

src/forms/upload/upload.component.js
StevePole: Would be useful to list the enums in the comment, rather than alternative versions ("file system", "live camera")

src/forms/upload/camera-only-upload/camera-only-upload.less
StevePole: I don't think I agree with that.  The ultimate destination is Neptune, but there is no plan to migrate styles from this repo to Neptune, whereas there is a plan to migrate Bootstrap styles there.  Until Neptune can provide an alternative to Bootstrap that can be consumed by our apps, I'd say Bootstrap is the right location.  All of the other upload styles are already there, we'd just be splitting the styles for a single component across multiple projects.

src/forms/upload/camera-only-upload/camera-only-upload.component.js
HaoYuanTW: Child needs to be able to change state of this variable in parent as parent is unware of when screen capture has completed.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
HaoYuanTW: planning to using $log, doing some testing though. All my debug logs are printing atm without me explicitly enabling debug log...

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        tryAcquireMediaStream(this)
+          .then((stream) => {
+            onVideoStreamAcquisition(this, stream);
+            this.CameraCaptureScreenHandler.getVideoSpecifications(this, isFullScreen);
+            this.$log.debug(`**screen resolved** : ${this.screenHeight} x ${this.screenWidth}`);
+          })
+          .catch((err) => {
+            this.$log.error(err);
```
StevePole: What happens for the user in this scenario?  I assume they're stuck, and do not see an error message?

HaoYuanTW: Need to transition out of upload flow to the other drop card and show an error msg. Im not sure if the current component allows for this error msg, need to investigate on this one.

src/forms/upload/camera-only-upload/camera-only-upload.less
HaoYuanTW: Would you mind if I keep the CSS here for the initial roll out? 
In that period I expect us to make small CSS changes to this component, screen size resolution/finding optimal button size and location can be a bit tricky. 
Keeping CSS here for a bit will help us to iterate easier.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,299 @@
...
+      }
+    });
+
+    this.displayCanvas = this.$element[0].querySelector('#video-preview #display-canvas');
+    this.uploadCanvas = this.$element[0].querySelector('#video-preview #upload-canvas');
+    this.videoPreviewElement = this.$element[0].querySelector('#video-preview');
+    this.video = this.$element[0].querySelector('#video-preview #video');
+    this.video.addEventListener('play', createVideoPlayCallback(this));
+
+    window.addEventListener('orientationchange', createOrientationChangeCallback(this), false);
```
StevePole: Rather than using `window` directly, inject angular's `$window` wrapper.  This enables you to mock the window for testing, including different screen sizes.  (There are further uses of window below, which you could also change)

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,299 @@
...
+        // After trying to acquire full screen, resolve video stream
+        // TODO haoyuan : firefox is recognizing pixel's btm bar
+        this.screenHeight = window.innerHeight;
+        this.screenWidth = window.innerWidth;
+        this.$log.debug(`screen : ${window.screen.height} x ${window.screen.width}`);
+        this.$log.debug(`screen available : ${window.screen.availHeight} x ${window.screen.availWidth}`);
+        this.$log.debug(`screen inner : ${window.innerHeight} x ${window.innerWidth}`);
+        this.$log.debug(`**screen resolved** : ${this.screenHeight} x ${this.screenWidth}`);
+
+        function onVideoStreamAcquisition($ctrl, stream) {
```
StevePole: No need to declare these function inline.  Can you move them out of the `finally` handler?

src/forms/upload/camera-only-upload/camera-only-upload.component.js
StevePole: Can we make that a callback instead? e.g. an `onCaptureComplete: '&'`?

demo/index.html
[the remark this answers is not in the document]
HaoYuanTW: Fixed model wiring, it wasnt working properly before, now it is :p

src/forms/upload/camera-only-upload/camera-capture-screen-handler.service.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+    inject(function($injector) {
+      $rootScope = $injector.get('$rootScope');
+      $compile = $injector.get('$compile');
+      $timeout = $injector.get('$timeout');
+      CameraCaptureScreenHandler = $injector.get('CameraCaptureScreenHandler');
+    });
+  });
+
+  it("gets correct video dimension for portrait narrow screen", function() {
```
StevePole: It's quite hard to follow why the code should behave as specified.  For example, when the screen height is 90, why should the video height be close to 111?  I have no idea... 

For tests to act as documentation they should describe the expected behaviour i.e. _why_ should it be 111?

I use the language of BDD to structure my tests and keep separation between test setup and expected results:
```
describe('when there is a narrow portrait screen', function() { 
  let videoWidth, videoHeight;
  beforeEach(function() {
    { videoHeight, videoWidth } = CameraCaptureScreenHandler.getVideoSpecifications(90, 40, 100, 40);
  };
  it('should set the video width to match the screen width', function() { // ????
    expect(videoWidth).toBeCloseTo(100, 2);
  });
  it('should set the video height to maintain the aspect ratio of the video resolution', function() { // ???
    expect(videoHeight).toBeCloseTo(111, 2);
  });
});
```
I've no idea what those expectation should actually be, but you get the idea, they should explain why those values are expected.

src/forms/upload/camera-only-upload/camera-capture-screen-handler.service.spec.js
```diff
@@ -0,0 +1,178 @@
+'use strict';
+
+describe('CameraCaptureScreenHandler', function() {
+  var $compile,
+    $rootScope,
+    $timeout,
+    CameraCaptureScreenHandler;
+
+  beforeEach(function() {
+    module('tw.styleguide.forms');
```
StevePole: Can you instantiate the module that you're actually testing here (I appreciate that this pattern wasn't followed elsewhere, I recently fixed this across the repo)

HaoYuanTW:
@StevePole 
Hey Steve, thank you for the massive help on this, I have addressed most of the concerns in code review and I have re-tested the updated component on iphone and 2 android devices. 

Pending issues are :
1. Moving CSS to bootstrap, can we ship the component as is for now, in the coming month and half I will likely need to do some adjustments, having CSS in one place will help me to make the adjustments faster. I want to move CSS over to bootstrap late October after 2-3 weeks of beta. 

2. Camera acquisition failure, I need to investigate more on what kind of errors can happen to build this better. It seems there is not a way for users to reset camera permissions without refreshing the page.

3. Android, this component sometimes look ugly on android because, for example, firefox on pixel was returning wrong innerwidth and height of browser screen, i cannot set video dimension correctly for it. 

4. UI, we want to go with current UI and start testing the component internally in terms of functionality and we have a 2 weeks window coming up where we can adjust the UI. I do not want to get to it now anymore due to time constraint on the project. 

With these being said, can we ship it now?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/styleguide-components",
-  "version": "3.10.3",
+  "version": "3.10.4",
```
StevePole: Adding new functionality is a minor release i.e. 3.11.0

src/forms/upload/upload.demo.html
```diff
@@ -99,36 +103,40 @@ <h4 id="upload">Upload</h4>
...
   button-text="{{$ctrl.buttonText}}"
   cancel-text="{{$ctrl.cancelText}}"
-<span ng-if="$ctrl.accept">
-  accept="{{$ctrl.accept}}"</span><span ng-if="$ctrl.wrongTypeText">
-  wrong-type-text="{{$ctrl.wrongTypeMessage}}"</span><span ng-if="$ctrl.maxSize">
+  accept="{{$ctrl.accept}}"
+  wrong-type-text="{{$ctrl.wrongTypeMessage}}"
   max-size="{{$ctrl.maxSize}}"
-  too-large-message="{{$ctrl.tooLargeMessage}}"</span><span ng-if="$ctrl.processingText">
-  processing-text="{{$ctrl.processingText}}"</span><span ng-if="$ctrl.successText">
```
StevePole: Why did you delete these spans?  They were there to keep the docs simpler when not using those options.

src/forms/upload/upload.controller.js
```diff
@@ -39,13 +44,31 @@ class UploadController {
     }
 
     this.addDragHandlers($scope, $element);
+
+    this.showLiveCaptureScreen = false;
+  }
+
+  onUploadButtonClick() {
+    if (this.isLiveCameraUpload) {
+      this.showLiveCaptureScreen = true;
```
StevePole: I don't think you need isLiveCameraUpload and showLiveCaptureScreen, they seem to do the same thing.  isLiveCameraUpload simply precalculates the value.  Because it's precalculated you need to do the $watch on $ctrl.source.  If you simply changed this line to:
```
this.showLiveCaptureScreen = this.source && this.source === 'CAMERA_ONLY';
```
I think you can achieve the same thing in a simpler way, with less code.

The other way way you use these is within the template.  I'll leave a separate comment there...

src/forms/upload/upload.demo.html
HaoYuanTW: The spans kept getting messed up when I was trying to add new parameters to the block and
I didnt think it was a big deal keeping these parameters even when they are not set.

src/forms/upload/upload.html
```diff
@@ -16,46 +17,54 @@
...
           ng-model="$ctrl.inputFile" />
-        <!-- ng-change="$ctrl.onManualUpload()" -->
       </label>
+
+      <!-- camera only upload -->
+      <tw-camera-only-upload
+        show-capture-screen="$ctrl.showLiveCaptureScreen"
+        on-capture-screen-close="$ctrl.onLiveCameraCaptureScreenClose()"
+        on-user-capture-confirmation="$ctrl.onFileConfirmation(file)"
+        ng-if="$ctrl.isLiveCameraUpload"
```
StevePole: Here we use isLiveCameraUpload && showLiveCaptureScreen in the interface to the component.  Again, I don't think we need both.  What's happening is that you use one to determine whether to include this component and another to trigger the component to start doing things.  You can simplify this by just having the component start doing things when it's compiled.  As long as you have this ng-if, it won't be included in the DOM, and it won't be initialised.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
```diff
@@ -0,0 +1,311 @@
...
+
+    if (!this.hasGetUserMedia()) {
+      // TODO: haoyuan how to handle get user media not being available?
+      this.$log.warn('getUserMedia() is not supported by your browser');
+    }
+
+    // entry point of this component
+    this.$scope.$watch('$ctrl.showCaptureScreen', (newValue, oldValue) => {
+      if (newValue !== oldValue && newValue) {
+        this.onLiveCamFlowStart();
```
StevePole: If you do the other things I suggest, you can remove this watch and just trigger this function straight away.  Basically the consumer will decide when to trigger the flow by including the component.  The logic already exists in the parent, it just does it by passing in a different prop.

I may also have made a bad suggestion earlier, it looks like this function is not a handler (reacting to the fact that something happened), but an imperative instruction (make this happen).  I would reserve `onXXX` for handlers, and rename this `startLiveCameraFlow()`.  Sorry!

src/forms/upload/upload.html
HaoYuanTW: >You can simplify this by just having the component start doing things when it's compiled

This is going to be used in mitigator section where we have multiple upload boxes and all of them are going to be initialized as source='CAMERA_ONLY', we cannot have all of them start rendering full screen video capture when they are initialized, hence we still need an explicit trigger.

src/forms/upload/camera-only-upload/camera-only-upload.spec.js
```diff
@@ -0,0 +1,181 @@
...
+          on-user-capture-confirmation='onUserCaptureConfirmation(file)'> \
+        </tw-camera-only-upload>";
+
+    // Create spies for callbacks
+    $scope.showCaptureScreen = false;
+    $scope.onCaptureScreenClose = jasmine.createSpy('onCaptureScreenClose');
+    $scope.onUserCaptureConfirmation = jasmine.createSpy('onUserCaptureConfirmation');
+
+    directiveElement = getCompiledDirectiveElement($scope, template);
+    directiveController = directiveElement.controller('twCameraOnlyUpload');
```
StevePole: We shouldn't interact with the controller of our component, it's an implementation detail of the component.  The consumers of our component do not interact with out controller, they interact via the component bindings (passing in data, responding to callbacks), and by interacting with the DOM (triggering clicks on template elements).  

We also don't care what controller functions are called, the only thing that makes a difference to our consumers are the external effects.  This focuses our tests on externally observable behaviours, and decouples our tests from specific implementation details that do not affect consumers (like internal function names).

These docs explain in much greater detail: https://transferwise.atlassian.net/wiki/spaces/EKB/pages/866583447/Testing+in+Angular

HaoYuanTW: too many interactions are required in this component where users need to click to accept things. I cannot test 2/3 of the interactions without mocking controller to bypass these interactions.

src/forms/upload/camera-only-upload/camera-only-upload.spec.js
```diff
@@ -0,0 +1,181 @@
...
+      spyOn(directiveController, 'tryAcquireFullScreen').and.returnValue($q.resolve());
+      spyOn(directiveController, 'tryAcquireMediaStream').and.returnValue($q.resolve());
+      spyOn(directiveController, 'onVideoStreamAcquisition').and.callFake(function() {});
+    });
+
+    it('Should try to acquire media stream after getting full screen', function() {
+      directiveController.tryAcquireFullScreen.and.returnValue($q.resolve());
+      $scope.showCaptureScreen = true;
+      $scope.$apply();
+      expect(directiveController.setScreenDimensions).toHaveBeenCalled();
```
StevePole: The better thing to test is the final dimension of the video in the template.  Right now, we don't test that setScreenDimensions behaves as expected (because it's a fake), and we don't test that this affects our template, which is what our user ultimately cares about.

src/forms/upload/camera-only-upload/camera-only-upload.spec.js
```diff
@@ -0,0 +1,181 @@
...
+  describe('After entering live upload flow', function() {
+    beforeEach(function() {
+      spyOn(directiveController, 'onLiveCamFlowStart').and.callFake(function() {
+        directiveController.onVideoStreamAcquisition(null);
+      });
+      $scope.showCaptureScreen = true;
+      $scope.$apply();
+    });
+
+    it('Should quit live upload flow when cancel is clicked', function () {
```
StevePole: Simulate the real behaviour instead
```
describe('when cancel is clicked', function() {
  beforeEach(function() {
    element.querySelector('.myCancelButton').dispatchEvent(new Event('click'));
  });
  it('should trigger the onCaptureScreenClosed handler', function() {
    expect($scope.onCaptureScreenClose).toHaveBeenCalled();
  });
});
```

StevePole: Sorry, I can see you did that actually.  The issue is that you check 'internal' effects(controller method calls), rather (or in addition) to 'external' effects (triggering of callback handlers, impact on DOM).

StevePole: If you separate out the trigger (the click) from the expectations (the outcomes) it will be clearer.

src/forms/upload/camera-only-upload/camera-only-upload.controller.js
HaoYuanTW: Changed the behavior here, if user failed to request to camera, we quit upload component. 

However we are not able to display a meaningful (translated) error msg asking users to refresh page to re-aquire camera permission due to limitation in this component that we can only supply one error msg.

src/forms/upload/camera-capture/camera-capture.component.js
```diff
@@ -0,0 +1,23 @@
+import controller from './camera-capture.controller.js';
+import template from './camera-capture.html';
+import './camera-capture.less';
+
+const CameraCapture = {
+  controller,
+  template,
+  bindings: {
+    cameraOverlay: '@', // Optional
+    cameraFaceMode: '@', // environment/user
```
StevePole: Could we rename these two 'overlay' and 'direction'?  I feel like the 'camera' context is clear given the nature of this component.

src/forms/upload/upload.component.js
```diff
@@ -31,7 +31,19 @@ const Upload = {
...
+
+    /* Source of file to upload
+     * Possible values
+     * CAMERA_ONLY (allow only camera upload)
+     * FILE_ONLY (allow only upload from file system)
+     * ANY (we dont care)
+     */
+    source: '<',
+    cameraOverlay: '@', // Optional
+    cameraFaceMode: '@' // environment/user
```
StevePole: Could we rename to cameraDirection?  (In this case the context is not clear, so the prefix makes sense)

src/forms/upload/upload.demo.html
```diff
@@ -140,6 +147,27 @@ <h5 class="page-header">
...
+        </label>
+        <tw-select
+          placeholder="File Source"
+          ng-model="$ctrl.source"
+          options="$ctrl.sources"></tw-select>
+      </div>
+
+      <div class="form-group" ng-init="$ctrl.size = 'md'">
+        <label class="control-label">
+          Camera Face Mode
```
StevePole: Camera Direction

src/forms/upload/upload.demo.html
```diff
@@ -140,6 +147,27 @@ <h5 class="page-header">
           class="btn btn-sm btn-default pull-xs-right"
           ng-click="$ctrl.makeFancy()">Make fancy</button>
       </h5>
+
+      <div class="form-group" ng-init="$ctrl.size = 'md'">
+        <label class="control-label">
+          File Source
```
StevePole: Is this the overlay?  Can we rename

src/forms/upload/camera-capture/camera-capture.component.js
HaoYuanTW: direction is not the former term for this, it should be facing mode, and dynamic interface has also adapted this convention, so it is consistent

PR #232, opened 2019-06-26
fix: divs with .error-min or .error-max are visible no matter the error

joebartels-tw:
the bug has the effect that for a given field:
```
{
  key: 'ssn',
  name: 'Your US Social Security Number',
  width: 'lg',
  type: 'text',
  required: true,
  hidden: false,
  placeholder: '111-22-3333',
  refreshRequirementsOnChange: false,
  displayFormat: '***-**-****',
  minLength: 9,
  maxLength: 9
}
```

the default `validationMessages` are being applied to it which include the [deprecated??] `minimum` and `maximum` keys.

and so the DOM nodes with classes `.error-minimum` or `.error-maximum` were not being hidden from UI

### DIFF
--- a/src/forms/field/field.controller.js
+++ b/src/forms/field/field.controller.js
@@ -1,3 +1,7 @@
+const STRING_VALIDATION_TYPES = ['async', 'required', 'min', 'max', 'pattern', 'minlength', 'maxlength'];
+const NUMBER_VALIDATION_TYPES = ['async', 'required', 'min', 'max'];
+const BOOLEAN_VALIDATION_TYPES = ['async', 'required'];
+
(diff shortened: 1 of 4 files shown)


src/forms/field/field.controller.js
```diff
@@ -1,3 +1,7 @@
+const STRING_VALIDATION_TYPES = ['async', 'required', 'pattern', 'minlength', 'maxlength'];
```
StevePole: date strings can utilise min and max

joebartels-tw:
fix already merged:
https://github.com/transferwise/styleguide-components/pull/237

PR #233, opened 2019-07-01
fix(select): filter duplicates based on label and value

pterblgh:
Comparing and eliminating duplicates in `twSelect` using only labels breaks most of the recipient dropdowns.

You can have several different recipients/bank account details using the same name - for example I can have 1000 recipients called `John Doe` but the current implementation would just allow me to render one in the available options.

This PR fixes that checking the values as well for the same label.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v3.8.5
+## Filter duplicates based on label and value in tw-select
+
 # v3.8.4
 ## Increase the tw-select large option size from 50 to 300
 
--- a/src/forms/select/select.controller.js
+++ b/src/forms/select/select.controller.js
@@ -134,23 +134,25 @@ class SelectController {
     }
 
     const filterStringLower =
-        this.filterString && escapeRegExp(this.filterString.toLowerCase());
+      this.filterString && escapeRegExp(this.filterString.toLowerCase());
 
-    const encounteredLabels = Object.create(null);
+    const encounteredLabelsAndValues = {};
 
     const filteredOptions = [];
     for (let i = 0; i < this.options.length; ++i) {
       const option = this.options[i];
 
-      const isDuplicate = encounteredLabels[option.label];
+      const isDuplicate =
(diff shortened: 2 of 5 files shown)


src/forms/select/select.controller.js
```diff
@@ -134,23 +134,35 @@ class SelectController {
...
+      const hasExistingValues = angular.isArray(existingValuesForLabel);
+
+      if (hasExistingValues) {
+        for (let j = 0; j < existingValuesForLabel.length; j++) {
+          if (angular.equals(existingValuesForLabel[j], option.value)) {
+            isDuplicate = true;
+            break;
+          }
+        }
+      }
```
sneakertack: [Subjective] Could also be expressed as:

```js
const isDuplicate = angular.isArray(existingValuesForLabel) &&
    existingValuesForLabel.some(existingValue => angular.equals(existingValue, option.value));
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# v3.8.5
+## Filter duplicates based on label and value in tw-select
```
sneakertack: If we're concerned people will upgrade and their use case breaks because we no longer deduplicate only on labels, how about considering a major version bump to v4? it's more likely that people upgrading from v3 to v4 will read the patch notes that way.

sneakertack:
:sweat_smile: :snail: Sad when you comment on a PR that has already been merged

src/forms/select/select.controller.js
pterblgh: I wanted to avoid any ES2015 Array methods because as far as I can see none of them except `.forEach()` is used in this project, so I didn't want to break this pattern.

CHANGELOG.md
pterblgh: I don't think this is a breaking change. The interface remained the same just the internal implementation changed which does not need any changes from consumers. 

The component's behaviour did change, I can see that, but I think this is still a patch version bump according to SemVer. No new features have been introduced in this PR.

PR #235, opened 2019-07-11
Empty card

laczikt:
Adds an empty state to the TW Card component, turning off the expand on click behavior.


StevePole:
What is the use case for this?  Seems a bit confusing to me, we're teaching our users a pattern for interaction, but then BAM! that pattern doesn't work any more.

PR #238, opened 2019-07-29
fix(select): display all headers passed to options

pterblgh:
Only the first header option was displayed due to a bug in the logic determining the filtered options.

The logic did not handle properly those option objects which have no label and value but only have `header` key.

This should now be fixed with this PR.

### DIFF
--- a/src/forms/select/select.spec.js
+++ b/src/forms/select/select.spec.js
@@ -1217,8 +1217,6 @@ describe('Select', function() {
   });
 
   describe('when multiple headers are passed', function() {
-    var $filterInput;
-
     beforeEach(function () {
       $scope.options = [
         { header: 'First' },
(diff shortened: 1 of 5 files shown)


src/forms/select/select.spec.js
```diff
@@ -1216,6 +1216,49 @@ describe('Select', function() {
...
+          <a href='' class='custom-action'> \
+              Custom action \
+          </a> \
+        </tw-select>";
+      component = getComponent($scope, template);
+      $filterInput = component.find(FILTER_INPUT_SELECTOR);
+    });
+
+    it('should show two header options', function() {
+      $filterInput.val("ca").trigger('change');
```
StevePole: Why do you need to trigger the filter to test this?

src/forms/select/select.spec.js
```diff
@@ -1216,6 +1216,42 @@ describe('Select', function() {
...
+          value: '1',
+          label: 'Cain'
+        }
+      ];
+      $scope.filter = 'Search';
+      var template = " \
+        <tw-select \
+          options='options' \
+          ng-model='ngModel' \
+          filter='{{filter}}'> \
```
StevePole: Don't think you need this line now, but not a big deal.

PR #239, opened 2019-08-06
Add support for persistAsync

StevePole:
twUpload can now post images immediately and extract an id from the response to bind to the model.  This can also be used by twFormControl and twField.

async-final


package.json
```diff
@@ -48,10 +48,10 @@
...
     "@uirouter/angularjs": "^1.0.22",
     "angular-mocks": "~1.5.0",
-    "babel-core": "^6.26.0",
-    "babel-eslint": "^7.2.3",
-    "babel-loader": "^7.1.2",
-    "babel-preset-es2015": "^6.24.1",
+    "babel-core": ">=6.26.0",
+    "babel-eslint": ">=7.2.3",
+    "babel-loader": ">=7.1.2",
+    "babel-preset-es2015": ">=6.24.1",
```
pterblgh: I think we should not use `>=` in devDependencies. Breaking changes can easily sneak into breaking the whole build pipeline.

src/forms/upload/async-file-saver.service.js
```diff
@@ -0,0 +1,46 @@
...
+  const $httpOptions = angular.copy($inputOptions);
+
+  if (!$httpOptions.url) {
+    throw new Error('You must supply a URL to post image data asynchronously');
+  }
+  if (!$httpOptions.headers) {
+    $httpOptions.headers = {};
+  }
+
+  $httpOptions.headers['Content-Type'] = undefined;
```
pterblgh: Can you please elaborate why do we need this?

src/forms/upload/async-file-saver.service.js
```diff
@@ -0,0 +1,46 @@
...
+
+    const $httpOptions = prepareHttpOptions(httpOptions);
+
+    const method = $httpOptions.method ? $httpOptions.method : 'POST';
+    delete $httpOptions.method;
+
+    // For testing
+    if (method === 'GET') {
+      return this.$http.get($httpOptions.url, $httpOptions);
+    }
```
pterblgh: Can we please remove this method related logic from the final service? We should never try to upload data using GET - if we want this to be really strict, we should throw an error here when the passed `$httpOptions` have a `GET` method property.

package.json
StevePole: Ah, crap, must have snuck back in a merge.

src/forms/upload/async-file-saver.service.js
StevePole: https://stackoverflow.com/questions/41607656/purpose-of-content-type-undefined-in-angularjs-fileupload

I'll add a comment in another PR.

src/forms/upload/async-file-saver.service.js
StevePole: The reason I had this was so that you try this from the demo page, where we just fetch a JSON file to represent the response.  Not sure of another easy way to do this short of a custom web server to provide that response?

pterblgh: You can override the service in the demo application by using the same injection token (name of the service) with a different implementation.

```js
class AsyncFileSaverDemoMock {
  save(fieldName, file, httpOptions) {
    // Do whatever needed here for demo to work
  }

angular.module('my.demo.module').service('AsyncFileSaver',  AsyncFileSaverDemoMock); 
```

If this code is executed after the original declaration of the `AsyncFileSaver`, you simple override the value referenced by the injection token and the injector will return your mocked instance in your demo application. 

This way the "production" service would contain only the necessary code for sending the request while the mocked service for demo purposes can implement the same interface and do whatever needed for the demo purposes.

I don't think this is hacky, as AngularJs works this way and will enable you to remove unnecessary code only for testing purposes from production code.

StevePole: Good idea, I wrote this before I extracted the AsyncFileSaver, when it would have been more challenging.

PR #240, opened 2019-08-08
Update dependencies

StevePole:
Remove jshint


package.json
```diff
@@ -44,46 +43,47 @@
...
+    "karma-firefox-launcher": ">=0.1.7",
+    "karma-intl-shim": ">=1.0.3",
+    "karma-jasmine": ">=2.0.1",
+    "karma-mocha-reporter": ">=2.2.4",
+    "karma-phantomjs-launcher": ">=1.0.0",
+    "karma-safari-launcher": ">=1.0.0",
+    "less": ">=3.9.0",
+    "live-server": ">=1.1.0",
+    "phantomjs-prebuilt": ">=2.1.16",
+    "release-to-github-with-changelog": ">=1.2.2",
```
pterblgh: I don't know if `>=` is the best approach, according to [NPM docs](https://docs.npmjs.com/misc/semver#ranges):

> For example, the comparator >=1.2.7 would match the versions 1.2.7, 1.2.8, 2.5.3, and 1.3.9, but not the versions 1.2.6 or 1.1.0.

Are you sure we want to enable auto bumping dependencies even some of those may contain breaking changes?

What I usually do when I'm updating multiple dependencies in a project is to delete the dependencies from `package.json` and do an `npm install dep1 dep2 dep3 dep4...` which is going to install the latest versions from the dependencies with a `^` versioning mechanism.

PR #249, opened 2019-09-18
Fix couple small JSON schema form bugs, Add demo form testing playground.

StevePole:
Bug affected schema validations when no ‘required’ array was present.

oneOfSchema component now defaults to the first schema that validates against the model (useful for selecting the correct schema when editing an existing model).

objectSchema component now does not change the input model, only broadcasts via onChange.

New test playground
image

Also modularises some demo code to separate files.

### DIFF
--- a/demo/playground.html
+++ b/demo/playground.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>JSON Schema Components - TransferWise</title>
+  <meta charset="UTF-8">
+  <meta name="viewport"
+    content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+
+  <script src="lib/jquery.min.js"></script>
+  <script src="lib/angular.min.js"></script>
+  <script src="lib/bootstrap.min.js"></script>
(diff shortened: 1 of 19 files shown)


demo/playground.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
```
oliverviljamaa: Amazing! Could we link to this playground from the docs and/or add it to the README?

src/json-schema/demo.controller.js
```diff
@@ -0,0 +1,81 @@
...
+  onModelChange(model, isValid, originatingShema) {
+    this.isValid = isValid;
+    console.log('Model changed', model); // eslint-disable-line
+    console.log('Is valid?', isValid); // eslint-disable-line
+    console.log('Triggered by schema', originatingShema); // eslint-disable-line
+    this.model = model;
+  }
+}
+
+Controller.$inject = ['$http'];
```
oliverviljamaa: Can we use class properties in this repo? If yes, we could move it to on top of the constructor, keeping the two places where we list those dependencies close together.

In addition, we could `export default class ...` directly.

PR #251, opened 2019-09-19
[ASIA-1973] Support new fields added for live camera upload component

HaoYuanTW:
Added support for new properties introduced to support camera capture flow (Dynamic interface already has them)

**uploadPlaceholderImage**
An image to replace icon in upload component to allow better customization
Screenshot 2019-09-20 at 15 43 39

**sourceType**
Source of file to upload, file system only or camera only or anything goes

**cameraGuideOverlay**
A guide to help users take better pictures. 
screenshot_1568731708 (1)

**cameraFacingMode** (camera direction)
Whether camera defaults to, say, selfie cam (front) or envronment cam (back)


HaoYuanTW:
@StevePole 
Hey Steve, you gave 2 additional points of actions last week
1. Refactor help properties into help-options
2. Refactor file source into upload-options

I have finished 1) but i do not wish to proceed with 2) here because upload-options is currently broken, base on git blame, for about 2 years? If I fix it now some designs in our website will change and I dont want to open a can of worms here.

src/forms/field/field.html
```diff
@@ -60,20 +61,20 @@
     {{ $ctrl.warningMessage }}
   </div>
 
-  <div ng-if="$ctrl.field.help"
+  <div ng-if="$ctrl.field.helpOptions"
```
StevePole: This was already correct, the name of the object is `help`.  The component might expose a prop, help-options, but that's the components concern, and doesn't need to match the definition in field exactly..

PR #254, opened 2019-09-27
Use jpeg with compression for camera upload

HaoYuanTW:
This change reduces captured image size by 10x

I have checked with verification team and they are fine with image resolution here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v5.3.2
+## Use JPEG compression for camera upload image
+We were using png before for camera uploads and the image was too big. This update reduces image size from camera upload component by up to 10x.
+
 # v5.3.1
 ## twSelect calls ngChange when the ngModel and ngRequired are set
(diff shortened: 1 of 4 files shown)


sneakertack:
A very welcome change, PNG's better suited for flat / vector-ish graphics, JPG's better for actual photos.

@xhook from Verification's also said OK to saving as JPEG on the initial capture. What's not okay is re-converting on the backend somehow.

https://transferwise.slack.com/archives/CQEN64U2E/p1585299544002900?thread_ts=1582215813.013000&cid=CQEN64U2E

> Read this thread a bit more carefully. You are compressing the image captured from the phone camera, right? In this case it’s alright.
Additional intermediate compression could be a problem in the case when you allow customer to upload a picture (regular upload) and then you recompress that (let’s say in the backend) with JPEG, then some interesting information may get lost. See https://en.wikipedia.org/wiki/Error_level_analysis for more info.

PR #256, opened 2019-10-02
Camera upload component bugfix

HaoYuanTW:
Camera direction parameter was not handled correctly causing camera direction to always default to 'environment'

### DIFF
--- a/src/forms/upload/camera-capture/camera-capture.controller.js
+++ b/src/forms/upload/camera-capture/camera-capture.controller.js
@@ -17,7 +17,10 @@ class CameraCaptureController {
     this.$q = $q;
     this.$window = $window;
     this.CameraCaptureScreenHandler = CameraCaptureScreenHandler;
+  }
 
+  $onInit() {
+    this.direction = this.direction || 'environment';
     this.cameraConstraints = {
       video: {
         width: {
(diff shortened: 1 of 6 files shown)


src/forms/upload/camera-capture/camera-capture.spec.js
```diff
@@ -42,6 +43,41 @@ describe('Given a camera capture component', function() {
     spyOn(controller, 'tryAcquireMediaStream').and.returnValue(acquireMediaStreamRequest.promise);
   });
 
+  describe('after initialization with camera direction', function() {
+    beforeEach(function() {
+      var template = " \
```
andrastothtw: These are miniscule details but why use `var` when you can use `const`? 🤔 In tests we should use the comfortable ES6 features.

src/forms/upload/camera-capture/camera-capture.controller.js
```diff
@@ -26,15 +29,11 @@ class CameraCaptureController {
           max: 1280
         },
         facingMode: {
-          ideal: this.direction
+          ideal: this.direction.toLowerCase()
```
andrastothtw: From this PR it is unclear to me if the value should be all lower case or all upper case. Here you lowercase it but everywhere else the values and labels are written in UPPERCASE. Is this on purpose? And if so, what's the purpose?

HaoYuanTW: facingMode.ideal : camera direction is the web standard and it requires lower case
however the values we pass in are from dynamic interface enum definition and they are in upper case

As camera direction property gets passed around our system, it should take the enum value which is in upper case until at the point they are consumed by a web component requiring conversion to lower case.

HaoYuanTW:
@andrastothtw Thank you for the review, i mean i created this bug so...

I made a tiny change based on your suggestion on const, are we good to go now? :)

andrastothtw:
@HaoYuanTW looks good to me!

PR #258, opened 2019-10-09
Feature update for camera upload component

HaoYuanTW:
1. Added camera overlay to guide user upload
Reason for need of this overlay this has been discussed extensively

Mobile Portrait
ezgif-1-e267514bad2e

Mobile Landscape
ezgif-1-ab5e4b389d00

Desktop
ezgif-1-f5beecff611f

2. Make selfie cam mirror user's action
We find that when overlay is not trivial (JPY's tilted overlay for example), if camera does not mirror the user, it is quite hard for user to put the ID doc inside the overlay. 

3. Default camera direction to selfie thus rendering the mirrored camera if system only has one video input (for desktop)
With changes in 2), camera in laptops will mirror the user by default, this is what we want.


HaoYuanTW:
Steve could see camera permissions pop up from his pixel. I need to double check that this thing works for android indeed.

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
```diff
@@ -107,6 +110,64 @@ class CameraCaptureScreenHandler {
...
+    this.$log.debug(`overlay width : ${width}`);
+    this.$log.debug(`overlay height : ${height}`);
+    this.$log.debug(`overlay x-offset : ${xOffset}`);
+    this.$log.debug(`overlay y-offset : ${yOffset}`);
+
+    return {
+      height,
+      width,
+      yOffset,
+      xOffset
```
StevePole: Wold it be cleaner to have two functions, one for getting the height/width, and another for getting the offsets?  If you end repeating a lot of logic, maybe not.  But this function is quite large at the moment.  I haven't reached the tests yet, but I would imagine this is very difficult to satisfactorily test.

src/forms/upload/camera-capture/camera-capture.html
```diff
@@ -1,11 +1,15 @@
...
-    <!-- TODO haoyuan: center and rotate camera overlay -->
-    <img id="overlay" ng-if="$ctrl.overlay" ng-show="$ctrl.showVideoInPreview" ng-src="{{$ctrl.overlay}}" alt="overlay" class="w-100 h-100 fixed-top opacity-overlay zindex-video-overlay"/>
-    <!-- TODO haoyuan: work with designer to figure out how this asset is going to be provided -->
-    <video id="video" ng-show="$ctrl.showVideoInPreview" ng-style="{height: $ctrl.videoHeight + '%', width: $ctrl.videoWidth + '%'}" class="fixed-top zindex-video" playsinline></video>
-    <canvas id="display-canvas" ng-show="$ctrl.showCaptureInPreview" class="fixed zindex-video" ></canvas>
+    <img id="overlay" ng-show="$ctrl.showVideoInPreview && $ctrl.overlayLoaded && $ctrl.videoPlaying"
+         ng-style="{left: $ctrl.overlayXOffset + 'px', top: $ctrl.overlayYOffset + 'px', width: $ctrl.overlayWidth + 'px', height: $ctrl.overlayHeight + 'px'}"
+         ng-src="{{$ctrl.overlay}}" class="fixed zindex-video-overlay"/>
+    <video id="video" ng-show="$ctrl.showVideoInPreview"
+           ng-style="{left: $ctrl.videoXOffset + '%', height: $ctrl.videoHeight + '%', width: $ctrl.videoWidth + '%'}"
```
StevePole: As these are complex I would do them with functions:
```
ng-style="$ctrl.getVideoStyles()"

function getVideoStyles() {
  return {
    left: this.videoXOffset + '%',
    height: ...
  };
}
```

src/forms/upload/camera-capture/camera-capture.html
```diff
@@ -1,11 +1,15 @@
...
   <!-- Video preview screen with the video itself and capture screen, should be full screen -->
   <div id="video-preview" ng-show="$ctrl.showVideoPreview" class="w-100 h-100 fixed-top bg-dark">
-    <!-- TODO haoyuan: center and rotate camera overlay -->
-    <img id="overlay" ng-if="$ctrl.overlay" ng-show="$ctrl.showVideoInPreview" ng-src="{{$ctrl.overlay}}" alt="overlay" class="w-100 h-100 fixed-top opacity-overlay zindex-video-overlay"/>
-    <!-- TODO haoyuan: work with designer to figure out how this asset is going to be provided -->
-    <video id="video" ng-show="$ctrl.showVideoInPreview" ng-style="{height: $ctrl.videoHeight + '%', width: $ctrl.videoWidth + '%'}" class="fixed-top zindex-video" playsinline></video>
-    <canvas id="display-canvas" ng-show="$ctrl.showCaptureInPreview" class="fixed zindex-video" ></canvas>
+    <img id="overlay" ng-show="$ctrl.showVideoInPreview && $ctrl.overlayLoaded && $ctrl.videoPlaying"
+         ng-style="{left: $ctrl.overlayXOffset + 'px', top: $ctrl.overlayYOffset + 'px', width: $ctrl.overlayWidth + 'px', height: $ctrl.overlayHeight + 'px'}"
+         ng-src="{{$ctrl.overlay}}" class="fixed zindex-video-overlay"/>
```
StevePole: I doubt ng-src needs these brackets, as most 'ng' prefixed attributes already expect an angular expression.

src/forms/upload/camera-capture/camera-capture.html
StevePole: Similar ones above and below.

src/forms/upload/camera-capture/camera-capture.less
```diff
@@ -119,14 +120,18 @@
...
-.opacity-overlay {
-  opacity: 0.05;
-}
-
 .zindex-video {
   z-index: 1030 !important;
 }
 
+.display-mirror {
+  -o-transform: scaleX(-1);
```
StevePole: Check browser compatibility I don't think these are necessary for our targets (ie10+)

src/forms/upload/upload.demo.html
```diff
@@ -168,6 +169,13 @@ <h5 class="page-header">
           options="$ctrl.cameraDirections"></tw-select>
       </div>
 
+      <div class="form-group" ng-if="$ctrl.source == 'CAMERA_ONLY'" ng-init="$ctrl.cameraOverlay = 'http://10.80.0.51:8887/jpn-id-edge-overlay.png'">
```
StevePole: What's this IP?  If you want to reference an image, you can place it in `/demo/images`

src/forms/upload/upload.demo.html
```diff
@@ -133,7 +133,8 @@ <h5>Configuration</h5>
   on-start="$ctrl.onStart"
   on-success="$ctrl.onSuccess"
   on-failure="$ctrl.onFailure"
-  on-cancel="$ctrl.onCancel"</span><span ng-if="$ctrl.source">
+  on-cancel="$ctrl.onCancel"</span><span ng-if="$ctrl.cameraOverlay">
+  overlay="{{$ctrl.cameraOverlay}}"</span><span ng-if="$ctrl.source">
```
StevePole: Are these braces {{ }} necessary?  I would have thought overlay was defined as a '>' prop, not '&'

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
```diff
@@ -49,13 +52,13 @@ class CameraCaptureScreenHandler {
...
   getCanvasSpecifications(
-    videoHeight, videoWidth,
+    videoHeightInPercentage, videoWidthInPercentage,
     screenHeight, screenWidth,
     videoResHeight, videoResWidth
   ) {
-    this.$log.debug('Recomputing canvas');
-    this.$log.debug(`video width : ${videoWidth}`);
-    this.$log.debug(`video height : ${videoHeight}`);
+    this.$log.debug('Computing canvas specs');
```
StevePole: Do we still need all the logs?

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
```diff
@@ -107,6 +110,64 @@ class CameraCaptureScreenHandler {
...
+
+  /**
+   * Overlay is going to be an image with appropriate width to indicate an ID
+   * If screen is portrait, leave a margin at both sides
+   * and strech overlay to fit screen width preserving ratio
+   * If screen is landscrape, set a fixed max width for overlay
+   * and stretch overlay to that width instead
+   *
+   * TODO : haoyuan make this more generic
+   * IMPORTANT: this method assumes an overlay of dimension roughly of an ID card
```
StevePole: Don't think this assumption is valid.  It shouldn't really matter what dimension the overlay has, we should always be able to choose an appropriate size.

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
StevePole: Moving this all to a service would also give give you a clean interface to test against.

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
HaoYuanTW: This is a constraint not an assumption. Activity and MC teams are working out some new ideas for overlay, and once they decide, web component should comply. 

In the current iteration of their ideas, we will be working with a fixed view port for camera, say, 1:1. This makes overlay rendering much easier and it poses constraint on overlay assets we accept. 

I dont want to think too hard how to make this one generic right now, it will become obsolete very soon.

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
```diff
@@ -107,6 +110,64 @@ class CameraCaptureScreenHandler {
...
+   * Overlay with height > width may not work
+   */
+  getOverlaySpecifications(
+    screenHeight, screenWidth,
+    overlayNaturalHeight, overlayNaturalWidth
```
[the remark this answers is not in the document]
HaoYuanTW: Natural height and width refers to the original height and width of image in an html image element. These names come from properties names in API doc below
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
HaoYuanTW: It would be good to have them...

src/forms/upload/upload.demo.html
HaoYuanTW: These are "@" props... Actually what is the rule here? I know "<" binding works for these properties too, but these are just string props, should they have been @ or < ?

HaoYuanTW:
Hey Steve, could you reconsider some of the ideas extracting overlay logic for now
1. Semantically the service is called CameraCaptureScreenHandler, overlay is a part of the camera capture screen
2. The methods in this handler are long right now, but mainly because we are handling 4-8 cases for each method pretty much - if you have a screen portrait/landscape and if the thing inside the screen is portrait/landscape. Not much computation there, majority of the code are repeating properties passed around. There are also some repeated common code in overlay logic such as getting screen orientation etc. 
3. height/width and offset are quite closely coupled, in some cases we handle offset first then get height and width based on offset restriction, in some cases, we get height and width first then compute the offset to render the height and width to the appropriate place on the screen. 
4. This design I have is quite different from what activity team and MC have in mind. I will change this very soon after they come to a conclusion. We are looking at 1-2 months tops.

src/forms/upload/upload.demo.html
StevePole: Either can work.  Generally if you expect something to be dynamic I would use '>'.  If it's something that's more likely fixed for the lifetime of the component, maybe a `size="small"` or it's a native HTML attribute, then I might use the string (@) variant.

HaoYuanTW: In this case the double braces will be needed there, the image url for overlay is defined as "@", it is static.

src/forms/upload/camera-capture/camera-capture-screen-handler.service.spec.js
```diff
@@ -101,9 +154,20 @@ describe('CameraCaptureScreenHandler', function() {
       expect(paintWidth).toBeCloseTo(90, 2);
       expect(paintHeight).toBeCloseTo(40, 2);
     });
+
+    it("gives overlay container mostly fitting the height and in upper middle portion of screen", function () {
```
StevePole: Now that you have a seperate service for calculating overlay size, I would mock it here, and just test that this component uses the overlay size provided by the service.  

Any tests to see how the overlay service behaves with different screen sizes or IDs can be done in the tests for the service.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# v3.12.5
```
StevePole: Feature enhancement bumps the middle number https://semver.org/

src/forms/upload/camera-capture/camera-capture-screen-handler.service.js
```diff
@@ -107,6 +112,64 @@ class CameraCaptureScreenHandler {
...
+   */
+  static getOverlayContainer(screenHeight, screenWidth) {
+    if (CameraCaptureScreenHandler.isScreenPortrait(screenHeight, screenWidth)) {
+      return {
+        height: screenWidth,
+        width: screenWidth,
+        yOffset: (screenHeight - screenWidth) / 2.5, // Place container above middle
+        xOffset: 0,
+      };
+    } else { // eslint-disable-line
```
StevePole: The reason eslint flags this is because your previous clause contain a return.  You can just do
```
if (something) {
  return ...
}
return ...
```
It will behave exactly the same.

HaoYuanTW: Hey Steve, yes i understand that, but the early return pattern is not always good

In this case, my "if" clause has strong logical relationship with "else" one indicating portrait, one indicating landscape. I didnt want to do early return intentionally, is this fine?

PR #263, opened 2019-11-15
AT-901 Temporarily remove httpOptions funcationality

ksunandp:
Temporarily disable, via comments, functionality attached to the httpOptions binding used in tw-upload. This functionality shouldn't be used anywhere in production yet, however when used, it causes issues regarding base url and blocks a user from continuing to submit.

Until an appropriate solution is found, we will have it commented out, not block work on backend and other clients.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/styleguide-components",
-  "version": "3.13.1",
+  "version": "4.0.0",
   "description": "TransfeWise AngularJS 1.x Styleguide Components",
   "license": "MIT",
   "scripts": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@transferwise/styleguide-components",
-  "version": "3.13.0",
+  "version": "3.13.1",
```
StevePole: Technically this is a breaking change...

ksunandp: But isn't the single use case of this feature (persistAsync) broken, and this a patch to stop it from breaking?

ksunandp: Upgraded to a major, I will change peer dependency in verification-components to allow only v4+ and get verification-components updated in all apps concerned.

PR #265, opened 2019-11-21
AT-924 New Service for Setting Config for Async Tasks

ksunandp:
## Context

We require a way to extend httpOptions for async functionality like persistAsync and validationAsync. In the example of using persistAsync for verification, we receive http options from the mitigator with no auth headers and a relative path for url. We need to have the option to extend those options by adding headers and/or prefixing the url with a baseUrl. Both optional for the consumer.
See https://transferwise.atlassian.net/wiki/spaces/DF/pages/909476078/Asynchronous+tasks for more details

## Changes

This PR creates a new service that can be injected by parent components to set base url and http headers and extend http options used in any async task (persistAsync/validationAsync). Setting these options is totally optional for the consumer.

## Considerations

The service is created to ensure there is no dependency on the consumer. And is only there to enhance the functionality when needed.
Current async functionality is not in use at the moment.


src/services/asyncTasksConfig/asyncTasksConfig.service.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  setHeader(header, value) {
+    this.headers[header] = value;
+  }
+
+  extendHttpOptions(inputOptions) {
+    const httpOptions = angular.copy(inputOptions);
+    httpOptions.headers = httpOptions.headers
+      ? angular.extend(httpOptions.headers, this.headers)
```
StevePole: I've started using the spread operator for this now.

src/services/asyncTasksConfig/asyncTasksConfig.service.js
```diff
@@ -0,0 +1,27 @@
+class AsyncTasksConfig {
+  constructor() {
+    this.apiUrl = '';
+    this.headers = {};
+  }
+
+  setApiUrl(apiUrl) {
```
StevePole: Can we be clearer that this is the baseUrl?  It could be either at the moment.

src/services/asyncTasksConfig/asyncTasksConfig.service.js
ksunandp: Yeah, originally used spread operator, however the repo doesn't have the correct babel plugin. Not sure if there's a reason for this, otherwise I'll just install.

src/forms/form-control/form-control.spec.js
```diff
@@ -11,6 +11,7 @@ describe('FormControl', function() {
 
   beforeEach(function() {
     angular.mock.module('tw.styleguide.forms.form-control');
+    angular.mock.module('tw.styleguide.services.async-tasks-config');
```
StevePole: Why is this necessary?

src/forms/upload/index.js
```diff
@@ -15,6 +16,7 @@ export default angular
   .service('AsyncFileSaver', AsyncFileSaver)
   .service('CameraCaptureScreenHandler', CameraCaptureScreenHandler)
   .service('CameraOverlayHandler', CameraOverlayHandler)
+  .service('AsyncTasksConfig', AsyncTasksConfig)
```
StevePole: We're registering AsyncTasksConfig as a service twice.  Once within it's own module and again here.  The way we should do it, is to list the module as a dependency of this module inside the `[]` on line 13 of this file.  I beleive onceyou do that, you can avoid explicitly mocking the `async-tasks-config` module within the tests.

src/forms/form-control/form-control.spec.js
StevePole: See below, I think I figured it out.

src/forms/field/field.spec.js
```diff
@@ -10,6 +10,7 @@ describe('Field', function() {
 
   beforeEach(function() {
     angular.mock.module('tw.styleguide.forms.field');
+    angular.mock.module('tw.styleguide.services.async-tasks-config');
```
StevePole: This one should also not be necessary.

src/forms/upload/index.js
ksunandp: Okay thanks! I did think there was something, but couldn't work it out. I will try that!

PR #267, opened 2019-11-26
AT896 happy path design for verification

johannacarlberg:
## Context

This is changing the design on the upload component if we go via persistAsync
link to task in jira: https://transferwise.atlassian.net/jira/software/projects/AT/boards/466?label=Web&selectedIssue=AT-896
Figma wireframes: https://www.figma.com/file/fjoFoudPJKTKtnBYcL5JpJ/ID-doc's?node-id=493%3A0

## Changes

When using persistAsync we have the new wireframes
- details and message will come back with the api response from persistAsync

## Considerations

This is only happy path, PR will follow with failed upload 

## Screenshots/GIFs

left is with persistAsync (httpOptions), right is old design and functionality

Screenshot 2019-11-26 at 11 40 39

### Before

see above 
### After

see above
## Checklist


src/forms/upload/upload.html
```diff
@@ -85,30 +82,57 @@ <h4 class="m-b-2" ng-if="$ctrl.label">
         <span class="icon icon-pdf icon-xxl" ng-if="!$ctrl.isImage"></span>
         <p class="text-ellipsis m-b-2">{{$ctrl.fileName}}</p>
       </div>
+      <div ng-if="!$ctrl.hasTranscluded && !$ctrl.isError && $ctrl.httpOptions && $ctrl.isImage" ng-style="{'background': 'linear-gradient(rgba(250, 250, 250, 0.9), rgba(250, 250, 250, 0.9)), url({{ $ctrl.image }}) no-repeat center', 'background-size':'contain'}">
```
ksunandp: On success, we should check for `successMessage` (see below). `httpOptions` may exist but they may be used differently or may not be used at all. So should stay with old flow.

src/forms/upload/upload.html
```diff
@@ -85,30 +82,57 @@ <h4 class="m-b-2" ng-if="$ctrl.label">
...
-            height="25">
-            <path d="M12 23a11 11 0 1 1 11-11 11 11 0 0 1-11 11zm0-21a10 10 0 1 0 10 10A10 10 0 0 0 12 2z"></path>
-            <path
-              d="M16.85 7.85l-.71-.7L12 11.29 7.85 7.15l-.71.7L11.29 12l-4.15 4.15.71.7L12 12.71l4.14 4.14.71-.7L12.7 12l4.15-4.15z">
-            </path>
-          </svg>
-          {{$ctrl.cancelText}}
-        </a>
-      </p>
+      <div ng-if="!$ctrl.httpOptions">
```
ksunandp: Same here as before, check if `successMessage` exists, otherwise show old flow. We can also probably put the `ng-if` on the `p` tag instead of creating another `div`.

src/forms/upload/upload.controller.js
```diff
@@ -268,6 +268,8 @@ function asyncSuccess(apiResponse, dataUrl, $ctrl) {
       && apiResponse.data[$ctrl.httpOptions.idProperty]) {
     const imageId = apiResponse.data[$ctrl.httpOptions.idProperty];
     $ctrl.setNgModel(imageId);
+    $ctrl.message = apiResponse.data.message;
+    $ctrl.details = apiResponse.data.details[0];
```
ksunandp: We should check details exists before getting first index. Just in case persistAsync is used for a different endpoint or things change.

PR #268, opened 2019-11-27
AT-897 persistAsync Feedback screen

ksunandp:
## Context

Feedback styles for when persistAsync returns a list of errors with the image.

## Changes

Shows a new screen regarding the feedback from the persistAsync call and allows for a modal for users to see the image and the errors associated.

### Before

Screenshot 2019-11-29 at 12 50 21

### After

Screenshot 2019-11-29 at 12 45 34
Screenshot 2019-11-29 at 12 45 42

## Checklist


src/forms/upload/upload.controller.js
```diff
@@ -295,10 +311,20 @@ function asyncSuccess(apiResponse, dataUrl, $ctrl) {
...
 
-function asyncFailure(error, $ctrl) {
+function asyncFailure(error, dataUrl, $ctrl) {
   // Start changing process indicator immediately
   $ctrl.processingState = -1;
 
+  if ($ctrl.httpOptions && error.data && error.data.message) {
+    $ctrl.errorMessage = error.data.message;
+    $ctrl.errorReasons = error.data.errors || [];
+    $ctrl.topError = $ctrl.errorReasons[0];
```
johannacarlberg: really picky 😱 , but could we rename this to like mainError or firstError? had to think a bit before i realised what topError could be...

ksunandp: I'll change to `firstError` 👍

PR #270, opened 2019-12-02
fix fieldset input width

johannacarlberg:
## Context

Before the fieldset used to have a class of btn-block (https://github.com/transferwise/styleguide-components/pull/266/files#diff-d07448f4be0dc1f1b6d4e37ce53fd3b0L53) when this was removed the fieldset was no longer 100%, see screenshots below. Hence adding 100% width to fix this. 

## Changes

Adding width: 100%  and flex:1 to fieldset to get full-width

### Before

Screenshot 2019-12-02 at 11 31 09
### After

Screenshot 2019-12-02 at 11 32 13

## Checklist

### DIFF
--- a/src/forms/fieldset/fieldset.less
+++ b/src/forms/fieldset/fieldset.less
@@ -1,4 +1,3 @@
-.full-width {
-    width: 100%;
+.tw-field-full-width {
     flex: 1;
 }
--- a/src/forms/fieldset/fieldset.less
+++ b/src/forms/fieldset/fieldset.less
@@ -1,0 +1,4 @@
+.full-width {
+    width: 100%;
+    flex: 1;
+}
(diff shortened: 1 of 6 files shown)


src/forms/fieldset/fieldset.less
```diff
@@ -0,0 +1,4 @@
+.full-width {
+    width: 100%;
+    flex: 1;
```
oleksandrbylim: I think flex:1 and width:100% is basically duplication, you need one or another
also if you'll decide to with flex, it should support cross-browser css properties

I believe bootstrap should contain something standard for this case

.also .full-width can sound as pretty common class, what 's about .tw-full-width or tw-field.full-width?

but please get at least two review from frontend guild channel

johannacarlberg: thanks i have updated

oleksandrbylim: pls show PR to somebody from #conversion

PR #278, opened 2020-02-04
AT-1097 Fix async action failure

ksunandp:
## Context

If an async action with upload, example persistAsync, fails. We don't want to halt the original flow or submitting the dataUrl. Only if the async action fails with a 422 with information to display to user, should be continue with the async action flow.

## Changes

Check error response for status code and continue as normal if not 422.

## Checklist

### DIFF
--- a/src/forms/upload/upload.spec.js
+++ b/src/forms/upload/upload.spec.js
@@ -336,19 +336,19 @@ describe('given an upload component', () => {
         expect(droppable.classList).toContain('droppable-complete');
       });
 
-      it('should show the failure message', function() {
+      it('should not show the failure message', function() {
         expect(directiveElement.querySelector('.upload-failure-message')).toBeFalsy();
       });
 
-      it('should not bind anything to the model', function() {
+      it('should bind base64url to the model', function() {
         expect($scope.ngModel).toBe(base64url);
       });
 
-      it('should not call the onSuccess handler', function() {
+      it('should call the onSuccess handler', function() {
         expect($scope.onSuccess).toHaveBeenCalled();
       });
 
-      it('should call the onFailure handler', function() {
+      it('should call not the onFailure handler', function() {
         expect($scope.onFailure).not.toHaveBeenCalled();
       });
     });
--- a/src/forms/upload/upload.spec.js
+++ b/src/forms/upload/upload.spec.js
@@ -322,9 +322,44 @@ describe('given an upload component', () => {
       });
     });
 
-    describe('when the timer has elapsed and the request was rejected', function() {
+    describe('when the timer has elapsed and the request failed', function() {
       beforeEach(function() {
-        deferred.reject({});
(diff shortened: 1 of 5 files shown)


src/forms/upload/upload.spec.js
```diff
@@ -322,9 +322,43 @@ describe('given an upload component', () => {
...
+
+      it('should show the failure message', function() {
+        expect(directiveElement.querySelector('.upload-failure-message')).toBeFalsy();
+      });
+
+      it('should not bind anything to the model', function() {
+        expect($scope.ngModel).toBe(base64url);
+      });
+
+      it('should not call the onSuccess handler', function() {
```
johannacarlberg: think this should be: should call the onSuccess handler

src/forms/upload/upload.spec.js
```diff
@@ -322,9 +322,43 @@ describe('given an upload component', () => {
...
+        expect($scope.onSuccess).toHaveBeenCalled();
+      });
+
+      it('should call the onFailure handler', function() {
```
johannacarlberg: and this should be: should not call the onFailure handler

src/forms/upload/upload.spec.js
```diff
@@ -322,9 +322,43 @@ describe('given an upload component', () => {
...
+      it('should show the complete screen', function() {
+        expect(droppable.classList).toContain('droppable-complete');
+      });
+
+      it('should show the failure message', function() {
+        expect(directiveElement.querySelector('.upload-failure-message')).toBeFalsy();
+      });
+
+      it('should not bind anything to the model', function() {
+        expect($scope.ngModel).toBe(base64url);
```
johannacarlberg: this should be: 'should bind'

src/forms/upload/upload.spec.js
```diff
@@ -322,9 +322,43 @@ describe('given an upload component', () => {
...
+
+      it('should not show the processing screen', function() {
+        expect(droppable.classList).not.toContain('droppable-processing');
+      });
+
+      it('should show the complete screen', function() {
+        expect(droppable.classList).toContain('droppable-complete');
+      });
+
+      it('should show the failure message', function() {
```
johannacarlberg: this should be 'should not show'

src/forms/upload/upload.spec.js
ksunandp: Oops (1/4)

src/forms/upload/upload.spec.js
ksunandp: Oops (2/4)

src/forms/upload/upload.spec.js
ksunandp: Oops (3/4)

src/forms/upload/upload.spec.js
ksunandp: Oops (4/4)

PR #280, opened 2020-02-12
WIP Add new multi-file upload component

StevePole:


gusso:
There's a gap between the line and where the padding starts, so the content of a row looks a bit misaligned.
Screenshot 2020-02-27 at 11 43 30 am

PR #282, opened 2020-02-27
Add new multi file upload component

luujam2:
## Context
For EDD, we want to allow the users to upload multiple files when we verify their source of funds. Currently the existing EDD flow only allows one file upload.

## Changes
This change introduces a multi file upload component.

## Considerations
- (PersistAsync) Currently no wiring of deletion of file on backend when the user removes a file in the UI
- (PersistAsync) We have left the logic where we fail only on 422. We will change this once the backend changes their persistAsync endpoint to save the file to ninjas in case the verification upload step fails.

## Screenshots/GIFs
upload

Screenshot 2020-02-27 at 11 27 35
Screenshot 2020-02-27 at 11 28 50
Screenshot 2020-02-27 at 11 29 09

## Checklist


gusso:
There's a gap between the line and where the padding starts, so the content of a row looks a bit misaligned.
Screenshot 2020-02-27 at 11 43 30 am

gusso:
- The container's left and right paddings should be the same as the one on the top, `24` instead of `32`.
- The margins between the row's elements should be `16` instead of `24`.
- The text style for the file status should be `.tiny` instead of `<small>`.
Frame 1

src/forms/upload/processing-card/processing-card.controller.js
```diff
@@ -110,17 +110,18 @@ function asyncSuccess(response, dataUrl, $ctrl) {
   $ctrl.$timeout(() => {
     $ctrl.isProcessing = false;
     $ctrl.isSuccess = true;
-  }, 3000);
+  }, 3700);
```
StevePole: Why do we need this change?  Not sure what impact it will have on the regular upload.

src/forms/upload/upload-button/upload-button.html
```diff
@@ -1,6 +1,9 @@
-<label class="btn btn-primary"
+<label class="btn btn-block btn-sm"
```
StevePole: I think this also change the single upload?

src/forms/upload/dropping-card/dropping-card.html
```diff
@@ -1,9 +1,9 @@
...
-  <h4 class="m-b-2">
-    {{ $ctrl.droppingText || 'Drop file to start upload' }}
-  </h4>
-  <div class="circle circle-sm">
+  <div class="circle circle-sm m-b-3">
     <span class="icon icon-add"></span>
   </div>
+  <h4>
+    {{ $ctrl.droppingText || 'Drop file to start upload' }}
+  </h4>
```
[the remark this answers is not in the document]
OFFLlNE: shouldn't this thing be translated?

src/forms/upload/upload-button/upload-button.html
luujam2: true, i'll add a condition in ng-class to add btn-sm if multiple is set to true

src/forms/upload/upload.demo.html
```diff
@@ -47,6 +47,7 @@ <h4 id="upload">Upload</h4>
     <div class="form-group">
       <tw-upload
         ng-model="$ctrl.modelToBind"
+        ng-change="$ctrl.log('test')"
```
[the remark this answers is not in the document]
luujam2: logging 'model change'

src/forms/upload/dropping-card/dropping-card.html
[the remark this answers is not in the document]
StevePole: When I was doing the preceding refactor I discovered that this string was never passed in.  I added a prop for it, but existing consumers won't have that prop.  I kept the existing behaviour (English text) as a fallback, although it's debatable whether we would be better to falling back to nothing.

src/forms/upload/processing-card/processing-card.controller.js
luujam2: discussed, changing the timeouts

src/forms/upload/multi-upload/multi-upload.demo.js
```diff
@@ -0,0 +1,79 @@
...
+
+  this.acceptOptions = [
+    { value: '.png', label: 'PNG (.png)' },
+    { value: '.jpg,.jpeg', label: 'JPG (.jpg,.jpeg)' },
+    { value: 'image/*', label: 'Images (image/*)' },
+    { value: 'video/*', label: 'Video (video/*)' },
+    { value: 'audio/*', label: 'Audio (audio/*)' }
+  ];
+
+  this.sizes = [
```
StevePole: Don't think you use this in your component.

src/forms/upload/multi-upload/multi-upload.html
```diff
@@ -0,0 +1,73 @@
...
+>
+  <h4 class="page-header text-xs-left m-r-3 m-l-3 m-b-0">{{$ctrl.label}}</h4>
+
+  <div class="droppable-dropping-card droppable-card">
+    <tw-upload-dropping
+      dropping-text="$ctrl.droppingText"
+    ></tw-upload-dropping>
+  </div>
+
+  <div ng-if="$ctrl.files.length === 0" id="empty-processing-list" class="droppable-default-card text-xs-center p-t-3">
```
StevePole: Can we avoid using 'id', what if two of these components are used on one page.  I know that would seem a bit weird, but technically it's possible

src/forms/upload/dropping-card/dropping-card.html
```diff
@@ -1,9 +1,9 @@
 <div class="droppable-card-content">
-  <h4 class="m-b-2">
-    {{ $ctrl.droppingText || 'Drop file to start upload' }}
-  </h4>
-  <div class="circle circle-sm">
+  <div class="circle circle-sm m-b-3">
```
StevePole: The processing card has m-b-2, this one has m-b-3, so I assume they do appear in the same position?

src/forms/upload/upload.less
```diff
@@ -25,3 +25,44 @@ div.transparent-area {
         min-height:72px;
     }
 }
+
+#multi-upload {
+    height: 375px;
+}
+
+#processing-list {
```
StevePole: Can we avoid using these ids for styling as well.  It should be possible to use more than one of these components, so they shouldn't figure ids, unless they are made programatically unique.

src/forms/upload/dropping-card/dropping-card.html
luujam2: yes, they are alligned

src/forms/upload/upload.less
luujam2: i'll change them to class names

src/forms/upload/upload.less
```diff
@@ -25,3 +25,44 @@ div.transparent-area {
         min-height:72px;
     }
 }
+
+#multi-upload {
```
StevePole: Any reason not to change this one?

PR #283, opened 2020-03-02
Minor fixes for upload component

HaoYuanTW:
## Context

I tested latest changes made here against send flow and discovered some problems
There is one console error i have addressed in the PR, also some subtle changes happened to angular digest. 

## Changes
Added asyncapply around some UI state changes
Fixed a miss routed parameter in callback

## Checklist

### DIFF
--- a/src/forms/form-control/form-control.controller.js
+++ b/src/forms/form-control/form-control.controller.js
@@ -36,9 +36,9 @@ class FormControlController {
   }
 
   // eslint-disable-next-line
-  onAsyncUploadSuccess(response) {
+  onAsyncUploadSuccess(data) {
     if ($ctrl.onAsyncSuccess) {
-      $ctrl.onAsyncSuccess({ response });
+      $ctrl.onAsyncSuccess(data.response);
     }
   }
 
--- a/src/forms/form-control/form-control.controller.js
+++ b/src/forms/form-control/form-control.controller.js
@@ -36,9 +36,9 @@ class FormControlController {
   }
 
   // eslint-disable-next-line
-  onAsyncUploadSuccess(data) {
+  onAsyncUploadSuccess({ file, response }) {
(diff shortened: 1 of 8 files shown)


src/forms/form-control/form-control.controller.js
```diff
@@ -36,9 +36,9 @@ class FormControlController {
   }
 
   // eslint-disable-next-line
-  onAsyncUploadSuccess(response) {
```
HaoYuanTW: response object here is a json containing response and also the file. therefore changing the wiring.

src/forms/upload/camera-capture/camera-capture.controller.js
```diff
@@ -397,7 +395,9 @@ function createUploadCallback($ctrl) {
     }
     $ctrl.showVideoPreview = false;
     $ctrl.closeVideoStream();
-    $ctrl.onCapture({ file: blob });
+    $ctrl.$scope.$apply(() => {
+      $ctrl.onCapture({ file: blob });
```
HaoYuanTW: without putting capture in an angular apply, this does not work in send flow. It works in demo app though.

src/forms/form-control/form-control.controller.js
HaoYuanTW: this is causing a console error in send flow during testing by the way

StevePole: I disagree with the way of fixing this, we should pass multiple arguments, not an object.

src/forms/form-control/form-control.controller.js
```diff
@@ -36,7 +36,7 @@ class FormControlController {
...
+  onAsyncUploadSuccess({ file, response }) {
```
HaoYuanTW: @StevePole like this?

PR #285, opened 2020-03-11
Remove 422 only failures for upload components

luujam2:
## Context
When we use persistAsync and the service returns a non 422 error, we should treat it as an error in the component instead of succeeding.

## Changes
Treat any error status codes as errors in the upload components

## Considerations
This will break the existing behaviour of ID checking since if the image upload fails with a 422, the user will not be able to carry on with the flow. This has been discussed and will only affect a small percentage of users

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v5.2.0
+# v5.1.2
 ## Remove failure for upload component on 422 only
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v6.0.0
+## Remove failure for upload component on 422 only
```
OFFLlNE: as this is a major version, please add the description why is this a breaking change and what the consumers should change to start consuming this new major version :D

PR #286, opened 2020-03-13
Display errors which are returned from persistAsync for multi file upload

luujam2:
## Context

Currently we only show the "failure-text" property if the upload fails. This is too vague.

## Changes
- Display the `tooLargeMessage` passed in by the consumer if the file is too large
- Provides an onFailure callback which exposes the error. an error message can be extracted and passed as failure-text back to the component
- Rename "secondaryButtonText" to "addMoreButtonText"

## Screenshots/GIFs
Failure due to file size:
Screenshot 2020-03-13 at 17 35 13
Failure due to server error:
Screenshot 2020-03-13 at 17 35 38
Failure where client can't extract error message from server:
Screenshot 2020-03-13 at 17 38 00

## Checklist

### DIFF
--- a/src/forms/upload/multi-upload/multi-upload.component.js
+++ b/src/forms/upload/multi-upload/multi-upload.component.js
@@ -14,7 +14,7 @@ const Component = {
     icon: '@', // illustration in icon shown in upload box
 
     buttonText: '@', // Button text shown in default state
-    secondaryButtonText: '@', // Button text shown in processing state
+    addMoreButtonText: '@', // Button text shown in processing state
 
     droppingText: '@', // Text shown when hovering with a file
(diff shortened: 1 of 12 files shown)


src/forms/upload/multi-upload/multi-upload.component.js
```diff
@@ -14,7 +14,7 @@ const Component = {
     icon: '@', // illustration in icon shown in upload box
 
     buttonText: '@', // Button text shown in default state
-    secondaryButtonText: '@', // Button text shown in processing state
+    addMoreButtonText: '@', // Button text shown in processing state
```
StevePole: comment is outdated

PR #288, opened 2020-03-17
add responseErrorExtractor callback for error message extraction

luujam2:
## Context

We want to display errors per file. The onFailure handler only modifies the root failureText which would change all error messages.

## Changes

Provides a responseErrorExtractor which is called when a file fails to upload. This function is passed the error object in which the message can be extracted and returned. The resulting error message is then assigned to `$ctrl.errorMessage` in `processing-card.controller.js`.

## Screenshots/GIFs
Screenshot 2020-03-17 at 17 44 03

Screenshot 2020-03-17 at 17 47 30

## Checklist

### DIFF
--- a/src/forms/upload/multi-upload/multi-upload.html
+++ b/src/forms/upload/multi-upload/multi-upload.html
@@ -51,6 +51,7 @@
         validation-messages="$ctrl.validationMessages"
         too-large-message="$ctrl.tooLargeMessage"
         cancel-text="$ctrl.cancelText"
+        response-error-extractor="$ctrl.responseErrorExtractor({error})"
       ></tw-upload-processing-mini>
     </div>
   </div>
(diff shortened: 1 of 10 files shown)


src/forms/upload/processing-card/processing.bindings.js
```diff
@@ -6,6 +6,7 @@ export default {
   onSuccess: '&',
   onFailure: '&',
   onCancel: '&',
+  responseErrorExtractor: '&',
```
StevePole: I would have done this slightly differently, using '<'.

You would also change the template, I'll leave another comment there.

src/forms/upload/multi-upload/multi-upload.html
```diff
@@ -51,6 +51,7 @@ <h4 ng-if="$ctrl.label" class="m-b-1">{{$ctrl.label}}</h4>
         validation-messages="$ctrl.validationMessages"
         too-large-message="$ctrl.tooLargeMessage"
         cancel-text="$ctrl.cancelText"
+        response-error-extractor="$ctrl.responseErrorExtractor({error})"
```
StevePole: As per other comment, this would become
`response-error-extractor="$ctrl.responseErrorExtractor"`

PR #296, opened 2020-04-21
Add checkbox-group component

gittymcgit:
As a follow up to this (https://transferwise.atlassian.net/wiki/spaces/DF/pages/1228965495/Multiple-choice+question)
## Context

We need to support checkbox groups in DF
## Changes

Added the component, also mapped to it in form-control
### Before
There was no such component

### After
image

## Checklist

### DIFF
--- a/src/forms/checkbox-group/checkbox-group.component.js
+++ b/src/forms/checkbox-group/checkbox-group.component.js
@@ -1,0 +1,20 @@
+
+import controller from './checkbox-group.controller.js';
+import template from './checkbox-group.html';
+
+const Checkbox = {
+  controller,
+  template,
+  require: {
+    $ngModel: 'ngModel'
+  },
+  bindings: {
+    name: '@',
+    ngModel: '=',
+    ngRequired: '<',
+    ngDisabled: '<',
+    options: '<',
+  }
+};
+
+export default Checkbox;
--- a/src/forms/checkbox-group/checkbox-group.component.js
+++ b/src/forms/checkbox-group/checkbox-group.component.js
@@ -10,6 +10,7 @@ const Checkbox = {
   },
   bindings: {
     name: '@',
+    ngChange: '&',
     ngModel: '=',
     ngRequired: '<',
     ngDisabled: '<',
--- a/src/forms/checkbox-group/checkbox-group.demo.html
+++ b/src/forms/checkbox-group/checkbox-group.demo.html
@@ -24,12 +24,10 @@
     <h5>Configuration</h5>
 <pre>&lt;tw-checkbox-group tw-validation
   ng-model="$ctrl.model"
+  options="$ctrl.options"
   ng-required="$ctrl.checkbox.required"
   ng-disabled="$ctrl.checkbox.disabled"
-  ng-click="$ctrl.log('checkbox click')"
-  ng-change="$ctrl.log('checkbox change')"
-  ng-blur="$ctrl.log('checkbox blur')"
-  ng-focus="$ctrl.log('checkbox focus')"</span>
+  ng-change="$ctrl.log('checkbox change')"</span>
   tw-validation&gt;&lt;/tw-checkbox-group&gt;</pre>
   </div>
   <div class="col-md-6">
--- a/src/forms/checkbox-group/checkbox-group.demo.html
+++ b/src/forms/checkbox-group/checkbox-group.demo.html
@@ -1,0 +1,66 @@
+<h4 id="checkbox-group">Checkbox Group</h4>
+
+<form>
+<div class="row m-b-3">
+  <div class="col-md-6">
+    <div class="form-group">
+      <label class="control-label">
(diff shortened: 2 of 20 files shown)


src/forms/checkbox-group/checkbox-group.component.js
```diff
@@ -0,0 +1,20 @@
...
+const Checkbox = {
+  controller,
+  template,
+  require: {
+    $ngModel: 'ngModel'
+  },
+  bindings: {
+    name: '@',
+    ngModel: '=',
+    ngRequired: '<',
```
StevePole, line 14: Can we add onChange here as well?  In dynamic forms, we don't rely on ngModel for changes, but use the onChange handlers.

src/forms/checkbox-group/checkbox-group.demo.html
```diff
@@ -0,0 +1,66 @@
...
+
+    <h5>Configuration</h5>
+<pre>&lt;tw-checkbox-group tw-validation
+  ng-model="$ctrl.model"
+  ng-required="$ctrl.checkbox.required"
+  ng-disabled="$ctrl.checkbox.disabled"
+  ng-click="$ctrl.log('checkbox click')"
+  ng-change="$ctrl.log('checkbox change')"
+  ng-blur="$ctrl.log('checkbox blur')"
+  ng-focus="$ctrl.log('checkbox focus')"</span>
```
StevePole, line 32: These don't exist :-)

src/forms/checkbox-group/checkbox-group.component.js
gittymcgit: I can add, but I don't think it's going to be used, since the components in DF are used through fieldset -> field -> form-control. So it only requires onchange in field and fieldset. Form control, that is going to use this component is using ngChange which should trigger on it's own when I update the model

StevePole: There's two things for me,  every other component provides on change, our components should present a consistent interface, so this should too.  

Every component in form-control works using ngChange, not by watching model changes, so you would need more change in FormControl:

https://github.com/transferwise/styleguide-components/blob/master/src/forms/form-control/form-control.html

gittymcgit: Oh okay. I think I understand what you mean now. What confused me was ngChange vs onChange. I think every other input component that is used in form-control has ng-change attribute, and then ngChange binding inside of it. Which then is not explicitly called in controller. Angular should call it when the model is updated, if I'm not mistaken. But I added that binding to my component now

StevePole: Sorry, my bad!

src/forms/field/field.controller.js
```diff
@@ -7,8 +7,15 @@ class FieldController {
     if (changes.initialField) {
       this.field = copyJSON(this.initialField);
 
-      this.control = this.field.control ? this.field.control
-        : this.RequirementsService.getControlType(changes.initialField.currentValue);
+      if (this.field.control === 'select' && this.field.selectType === 'CHECKBOX') {
+        this.control = 'checkbox-group';
```
StevePole, line 11: Where does this 'selectType' come from, is that new?   Checkbox isn't a type of select box, it's a different type of control, so I would pass it in as such.

As much as possible we should be using json schema internally, if we need to handle 'legacy' formats, we should do that using the requirements service.  In JSON schema I would expect this to appear as:
```
{
  type: "array",
  items: {
    enum: [ 1, 2 ],
    values: [{value: 1, label: "One"}, {value: 2, label: "Two"}
  }
}
```
We will likely translate that into an explicit instruction to render a checkbox group based on the size of the array of enums.

If we want to provide an override of the normal logic, we would include that in the schema:
```
{
  type: "array",
  control: "checkbox-group",
  items: {
    enum: [ 1, 2 ],
    values: [{value: 1, label: "One"}, {value: 2, label: "Two"}
  }
}
```
Though `checkbox-group` may not be the best name, as on mobile it might be those toggle things they have...

src/forms/fieldset/fieldset.demo.js
```diff
@@ -227,6 +227,24 @@ function fieldsetDocsController() {
           }
         ]
       },
+      checkboxGroup: {
+        title: 'Checkbox Group',
+        type: 'string',
+        control: 'select',
+        selectType: 'CHECKBOX',
```
StevePole, line 234: We need to update these

src/forms/fieldset/fieldset.demo.js
```diff
@@ -227,6 +227,24 @@ function fieldsetDocsController() {
...
+        width: 'md',
+        refreshRequirementsOnChange: true,
+        values: [
+          {
+            key: '1',
+            name: 'One'
+          },
+          {
+            key: '2',
+            name: 'Two'
```
StevePole, line 244: These are an old format as well

PR #299, opened 2020-05-19
Change validation approach, support control descriptions and show select placeholder even when required

StevePole:
[Note: due to the addition of prettier, and the need to change a lot of files, a lot of additional lines were edited, sorry!]

## Context
In some circumstances the select placeholder provides valuable information that the customers need to see.  In other scenarios we provide a description containing useful information.

## Changes
We currently hide select placeholders from the opened dropdown when the field is required.  We will now show it as an unclickable entry.  

We can now show descriptions against checkboxes and radio, these are added to the docs.

We switch the validation to follow the same approach as our React implementation, this allows us to handle the schema validation and decisions on when to show error messages, rather than relying on Angular's ngModel validation pipeline.

### Before
image

### After
image

image

## Checklist


src/forms/fieldset/fieldset.demo.html
```diff
@@ -6,15 +6,18 @@ <h2 id="fieldsets" class="page-header">
...
+    </p>
+    <p>
+      Fieldsets provide a 2-way bound model that is updated as the user interacts. There is also a
+      2-way bound 'is-valid' property, that can be useful for enabling/disabling submit buttons.
+    </p>
   </div>
   <div class="col-md-6">
     <h5>Component API</h5>
-<pre>
+    <pre>
```
giulioleso, line 20: minor: not sure if it's GitHub but seems the indentation is off here

giulioleso:
It seems all ok for me, but before approving it I'd prefer to have another pair of eyes on it

src/forms/fieldset/fieldset.demo.html
StevePole: It's prettier doing the formatting.  It's technically correct now, as the <pre> is a child of the div[class=col-md-6]. The following lines are indented differently because the browser respects white space inside a 'pre', so the content cannot be indented in the same way. Before we were kind of 'grouping' the pre with it's content, rather than grouping the 'pre' with it's HTML siblings.

src/forms/select/select.html
```diff
@@ -76,29 +92,49 @@
...
-      <a href="" tabindex="-1"
+    <li
+      ng-if="$ctrl.placeholder && !$ctrl.filter"
+      ng-class="{
+        active: !$ctrl.selected && !$ctrl.ngRequired,
+        'dropdown-header': $ctrl.ngRequired
+      }"
+    >
+      <span ng-if="$ctrl.ngRequired" class="text-ellipsis" ng-click="$event.stopPropagation()">
+        {{$ctrl.placeholder}}
```
helendurranttw, line 126: Why would we not show the placeholder if it's not required?

helendurranttw: Oh. I should have read the next line. :)

src/json-schema/basic-type-schema/controller.js
```diff
@@ -1,25 +1,61 @@
...
+    const validModel = getValidModelParts(model, this.schema);
 
-    if (this.onChange) {
-      this.onChange({ model, schema: this.schema });
+    this.validationKeys = getValidationFailures(validModel, this.schema, this.required);
+
+    const broadcastModel = this.validationKeys.length ? null : validModel;
+
+    if (!isNull(broadcastModel)) {
+      this.internalModel = model;
```
helendurranttw, line 43: Just checking, at this point if validation passes, will validModel and model be equivalent? You don't want to be setting this to broadcastModel instead of model?

src/json-schema/basic-type-schema/template.html
```diff
@@ -1,13 +1,16 @@
...
   locale="{{ $ctrl.locale }}"
   on-change="$ctrl.onModelChange(value)"
   error-message="$ctrl.errors"
   warning-message="$ctrl.schema.help.warning"
   validation-messages="$ctrl.getValidationMessages()"
   upload-options="$ctrl.translations.upload"
-  class="btn-block"> <!-- Remove btn-block after boostrap update -->
+  class="btn-block"
+>
+  <!-- Remove btn-block after boostrap update -->
```
helendurranttw, line 15: 😢

src/json-schema/validation/rule-validators/index.js
```diff
@@ -1,14 +1,17 @@
...
+} from '../type-validators';
 
 function isValidRequired(value, isRequired) {
-  return !isRequired || typeof value !== 'undefined';
+  return !isRequired || !isUndefined(value);
 }
 
 function isValidMinLength(value, minLength) {
-  return typeof minLength === 'undefined' || (!!value && value.length >= minLength);
+  return !isNumber(minLength) || (isString(value) && value.length >= minLength);
```
helendurranttw, line 10: 😌

src/json-schema/validation/rule-validators/index.js
```diff
@@ -20,29 +23,35 @@ function isValidPattern(value, pattern) {
   }
 }
 
-function isValidMax(value, max) {
-  return typeof max === 'undefined' || (typeof value !== 'undefined' && value <= max);
+function isValidMaximum(value, maximum) {
+  return (
+    (!isNumber(maximum) && !isString(maximum))
```
helendurranttw, line 28: How is this supposed to work for strings values, or string maximums?

src/json-schema/validation/rule-validators/index.js
```diff
@@ -20,29 +23,35 @@ function isValidPattern(value, pattern) {
...
+    || ((isNumber(value) || isString(value)) && value <= maximum)
+  );
 }
 
-function isValidMin(value, min) {
-  return typeof min === 'undefined' || (typeof value !== 'undefined' && value >= min);
+function isValidMinimum(value, minimum) {
```
helendurranttw, line 33: Same question as the above

src/forms/select/select.html
StevePole: Tbf, this optimisation may be a bit overkill...

src/json-schema/validation/rule-validators/index.js
StevePole: '2000-01-02' > '2000-01-01'

JSON (& therefore JSON schema) doesn't have a date type, only `type: "string"` and `format: "date"`, therefor it doesn't have minDate and maxDate validations, only minimum/maximum. Because the dates are iso8601 they work as pure string comparisons.

src/json-schema/validation/type-validators/index.js
```diff
@@ -1,9 +1,8 @@
-
 function isString(value) {
   return typeof value === 'string';
 }
 function isNumber(value) {
-  return typeof value === 'number';
+  return typeof value === 'number' && !isNaN(value); // eslint-disable-line
```
helendurranttw, line 5: Is the eslint issue here about global `isNaN`? You can probably use `Number.isNaN` here - difference is `Number.isNaN` checks for the explicit `NaN` primitive, but `isNaN` will just check if something is not a number - so `isNaN('blergh')` will return true but `Number.isNaN('blergh')` will return false.

src/json-schema/validation/validation-failures/index.js
```diff
@@ -101,8 +118,30 @@ function getBooleanValidationFailures(value, schema, isRequired) {
...
+  return [];
+}
+
+function getConstValidationFailures(value, schema, isRequired) {
+  if (!isValidRequired(value, isRequired)) {
+    return ['required'];
+  }
+
+  if (!isNull(value) && value !== schema.const) {
+    return ['enum'];
```
helendurranttw, line 138: Should this be const?

src/json-schema/validation/rule-validators/index.js
helendurranttw: What about other strings? Do we need to protect against things like '4' > '05'?

src/json-schema/validation/validation-failures/index.js
StevePole: Good spot.  It is slightly debatable, `const: 1` is syntactic sugar for `enum: [1]`, so when it comes to whether this is an incorrect type, we could portray it either way.  It may be better to give the consumer the opportunity of a more specific error message 🤔

src/json-schema/validation/rule-validators/index.js
StevePole: Depends what you mean by 'protect', there's no intrinsic reason not to allow that comparison, though it would be an odd thing.  It might be that we use the backend library to prevent people from specific minimum/maximum unless we're dealing with a date format.

src/json-schema/validation/type-validators/index.js
StevePole: Interesting, it was there just because of the global.  `typeof NaN === 'number'` so we needed the extra check.

src/json-schema/basic-type-schema/controller.js
StevePole: We don't want to set it to the broadcastModel, because doing so would wipe invalid user entered values from the control.  So because we have managed input components we need the invalid value to maintain the view, hence the original model is the thing to use.

src/json-schema/validation/rule-validators/index.js
helendurranttw: Right - I think that would make sense.

PR #300, opened 2020-05-22
Update camera component to follow the Camera Guidelines spec

sneakertack:
## Context
When this camera component is used in Dynamic Forms, it expects the outline image to be provided by `camera.overlay` on the field. This was true at the start, **but since then `camera.overlay` has evolved to mean a different thing, and the outline should be provided on `camera.outline` instead.**

Android and iOS Dynamic Flows are already following the new convention, so this PR updates the web client to align with them.

Some Slack discussion: https://transferwise.slack.com/archives/C5FBXBX8W/p1590050035160900

## Changes

1. Update the overlay logic to use the `outline` field.
    - But to preserve backward compatibility (useful during the migration), we'll still let it work with `overlay` for now. This should be removed once the producers (namely Mitigator for Japan) have been corrected to always return `outline`.
2. [Internal] Wrap everything up in a `camera-guidelines` prop, instead of passing individual props like `camera-overlay` and `camera-direction`.
    - So the new prop just gets passed untouched from field -> form-control -> upload -> capture-card -> camera-capture, and the camera-capture subcomponent is the one that really uses it.
    - This will make things easier when we want to start supporting the rest of the new camera-guidelines features too (like `overlay`, `mesh`, `instructions` etc.)

## Considerations
Mostly covered above. With this PR we won't fully support whatever is specified by Camera Guidelines but we're 1 step closer to getting there.

Searched for whether any external user is using the outgoing `camera-options` prop, couldn't find any external users, so the rename should be safe: https://github.com/search?q=org%3Atransferwise+%22camera-options%22&type=Code

Demo has been updated accordingly:

Untitled

## Checklist


src/forms/field/field.html
```diff
@@ -35,7 +35,7 @@
     validation-async="$ctrl.field.validationAsync"
     persist-async="$ctrl.field.persistAsync"
     file-upload-source="$ctrl.field.sourceType"
-    camera-options="$ctrl.field.camera"
+    camera-guidelines="$ctrl.field.camera"
```
StevePole, line 38: I don't think we need to change the property name?  Would prefer not to put a breaking change on the component interface unless we have to.

src/forms/form-control/form-control.html
```diff
@@ -94,8 +94,7 @@
 
       http-options="$ctrl.persistAsync"
       source="$ctrl.fileUploadSource"
-      camera-overlay="{{$ctrl.cameraOptions.overlay}}"
-      camera-direction="{{$ctrl.cameraOptions.direction}}"
+      camera-guidelines="$ctrl.cameraGuidelines"
```
StevePole, line 97: Changing this one makes more sense, and I consider the 'FormControl' more of an implementation detail of the Field.

src/forms/upload/camera-capture/camera-capture.controller.js
```diff
@@ -22,6 +22,21 @@ class CameraCaptureController {
   }
 
   $onInit() {
+    this.guidelines = this.guidelines || {};
+
+    // Backward compatibility: older producers (e.g. Japan live photo IDs from Mitigator)
```
StevePole, line 27: We also have 'RequirementsService' where we port older formats to the latest spec.  I did it this way so that the implementation can focus only on latest formats.  That's probably a better way to go.

sneakertack: Got it :+1:, addressed with https://github.com/transferwise/styleguide-components/pull/300/commits/cbf0423df7e5214fc71b96fb83b5d37bb8534557

PR #305, opened 2020-06-15
Update Australian External Territories entry in telephone countries

fergusjordantw:
## Context

https://transferwise.atlassian.net/browse/SEND-737

When creating a profile in the transferFlow we pass in the users locale to twTelephone based on the country they registered with. In the case of Australia, this is `en-AU`, which currently prefills the telephone input with `+672` - the phone code for Antarctica. 

This is because it's getting multiple hits for the `AU` `iso2` lookup it's doing in `countries.json`, and it is (for whatever reason) preferring whichever phone code is longer https://github.com/transferwise/styleguide-components/blob/master/src/forms/telephone/telephone.controller.js#L143-L149

## Changes

Updates the Australian External Territories entry in `countries.json` to use the Antarctica iso2 code, supported by this note in Wikipedia on Australias external territories https://en.wikipedia.org/wiki/ISO_3166-2:AU#External_territories

image

## Considerations

- Not an expert on the ISO spec and found it very hard to find a legitimate source of truth for this country code as it's a bit of an outlier and isn't explicitly mentioned on their website under AU or AQ: https://www.iso.org/obp/ui/#iso:code:3166:AQ https://www.iso.org/obp/ui/#iso:code:3166:AU
- If we do go forward with updating the iso2 to `AQ` instead of `AU`, does it follow that we should update the iso3 to `ATA` instead of `AUS`?
- My test only implicitly covers this issue, is it worth adding an explicit test for this use-case?

## Checklist

### DIFF
--- a/src/forms/telephone/countries.json
+++ b/src/forms/telephone/countries.json
@@ -76,7 +76,7 @@
 }, {
 	"name": "Australian Antarctic Territory",
 	"iso2": "AQ",
-	"iso3": "AUS",
+	"iso3": "ATA",
 	"phone": "+672"
 }, {
 	"name": "Austria",
--- a/src/forms/telephone/countries.json
+++ b/src/forms/telephone/countries.json
@@ -74,8 +74,8 @@
 	"iso3": "AUS",
 	"phone": "+61"
 }, {
-	"name": "Australian External Territories",
-	"iso2": "AU",
+	"name": "Australian Antarctic Territory",
+	"iso2": "AQ",
 	"iso3": "AUS",
 	"phone": "+672"
 }, {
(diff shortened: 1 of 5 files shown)


src/forms/telephone/countries.json
```diff
@@ -74,8 +74,8 @@
 	"iso3": "AUS",
 	"phone": "+61"
 }, {
-	"name": "Australian External Territories",
-	"iso2": "AU",
+	"name": "Australian Antarctic Territory",
+	"iso2": "AQ",
 	"iso3": "AUS",
```
Tankenstein, line 79: I'd update this to `ATA` as well, as far as i understand. This way at least we have internal consistency.

PR #307, opened 2020-07-09
Make checkbox-group model to be an array

gittymcgit:
## Changes

Moving checkbox-group model conversion into string into fieldSet component as it will only be required for v2
## Checklist

### DIFF
--- a/src/forms/field/field.html
+++ b/src/forms/field/field.html
@@ -18,7 +18,7 @@
     upload-accept="{{ $ctrl.field.accept }}"
     upload-icon="{{ $ctrl.field.icon }}"
     upload-too-large-message="{{ $ctrl.field.tooLargeMessage }}"
-    options="$ctrl.field.values ||$ctrl.field.items.values"
+    options="$ctrl.getOptions()"
     upload-options="$ctrl.uploadOptions"
     ng-model="$ctrl.model"
     ng-focus="$ctrl.onFocus()"
--- a/src/forms/field/field.html
+++ b/src/forms/field/field.html
@@ -18,7 +18,7 @@
     upload-accept="{{ $ctrl.field.accept }}"
     upload-icon="{{ $ctrl.field.icon }}"
     upload-too-large-message="{{ $ctrl.field.tooLargeMessage }}"
-    options="$ctrl.field.values"
+    options="$ctrl.field.values ||$ctrl.field.items.values"
     upload-options="$ctrl.uploadOptions"
     ng-model="$ctrl.model"
     ng-focus="$ctrl.onFocus()"
--- a/src/forms/fieldset/fieldset.controller.js
+++ b/src/forms/fieldset/fieldset.controller.js
@@ -11,6 +11,7 @@ class FieldsetController {
     if (!this.model) {
       this.model = {};
     }
+    this.internalModel = this.parseModel();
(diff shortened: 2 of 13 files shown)


src/forms/field/field.html
```diff
@@ -18,7 +18,7 @@
     upload-accept="{{ $ctrl.field.accept }}"
     upload-icon="{{ $ctrl.field.icon }}"
     upload-too-large-message="{{ $ctrl.field.tooLargeMessage }}"
-    options="$ctrl.field.values"
+    options="$ctrl.field.values ||$ctrl.field.items.values"
```
StevePole, line 21: Can we encapsulate this in a function that makes the decision based on the field type i.e. when it's an array, pass items.values, else pass values?

src/forms/fieldset/fieldset.controller.js
```diff
@@ -50,6 +52,27 @@ class FieldsetController {
     }
   }
 
+  parseModel() {
+    const parsedValues = {};
+    Object.keys(this.fields).forEach((key) => {
+      if (this.fields[key].control === 'checkbox-group' && this.model && typeof this.model[key] === 'string') {
```
StevePole, line 58: I'm wondering if 'this.model' is safe here?  What if we were dealing with an enum: [0, 1, 2]?  The 0 would fail.

StevePole: Maybe we're going in the other direction where that isn't a concern? 🤔

StevePole: Yeah, I think it's fine.

src/forms/fieldset/fieldset.controller.js
```diff
@@ -77,6 +97,15 @@ class FieldsetController {
 
     // Delay so model can update
     this.$timeout(() => {
+      this.model = this.stringifyObjectsInModel();
+      if (this.onFieldChange) {
+        if (field && field.control === 'checkbox-group') {
+          this.onFieldChange({ value: JSON.stringify(value), key, field });
```
StevePole, line 103: Why does this need to move from its original location?

src/forms/fieldset/fieldset.demo.js
```diff
@@ -229,7 +229,7 @@ function fieldsetDocsController() {
       },
       checkboxGroup: {
         title: 'Checkbox Group',
-        type: 'string',
+        type: 'array',
```
StevePole, line 232: This might need some management in the RequirementsService, I'm not sure what is going to come from the API, it may be that we need to map 'string' -> 'array'?

src/forms/fieldset/fieldset.controller.js
gittymcgit: I think since it's gonna be in array []  it should be fine. [0] should be truthy

src/forms/fieldset/fieldset.controller.js
gittymcgit: For nested components to work as before, we need to allow the model to change first, and only then trigger events, otherwise the change won't propagate

src/forms/fieldset/fieldset.demo.js
gittymcgit: yep, I'm doing that in fieldset

PR #309, opened 2020-07-14
Make Fieldset remove invalid properties when it receives new fields

StevePole:
## Context
Sometimes when we do a refresh of the form, we end up with different fields, in that case we should remove any properties from the model that no longer appear in the set of fields

## Changes
Alongside those changes we also improve the checkbox-group implementation to work with numeric enums as well as strings.  The requirements service now better identifies checkbox-groups based on their data type, and we have improved the definition list to show the labels for the enums rather than the raw values.

## Screenshots/GIFs
image
image

## Checklist

### DIFF
--- a/src/forms/checkbox-group/checkbox-group.controller.js
+++ b/src/forms/checkbox-group/checkbox-group.controller.js
@@ -1,85 +1,98 @@
 
 class CheckboxGroupController {
-  constructor($scope, $element, TwDomService, $timeout) {
-    const $ngModel = $element.controller('ngModel');
+  constructor($element, TwDomService) {
+    this.$ngModel = $element.controller('ngModel');
 
     this.dom = TwDomService;
     this.$element = $element;
-    this.element = $element[0];
-    this.internalModel = {};
-    this.$timeout = $timeout;
 
-    this.addWatchers($scope, $element, $ngModel);
-    $ngModel.$render = () => this.updateInternalValue($ngModel);
+    this.internalModel = [];
+    this.internalOptions = [];
+  }
+
+  $onChanges(changes) {
+    if (changes.options) {
+      this.onExternalOptionsChange(changes.options.currentValue);
+    }
+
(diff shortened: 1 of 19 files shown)


src/forms/checkbox-group/checkbox-group.controller.js
```diff
@@ -1,85 +1,98 @@
...
+      this.onExternalModelChange(changes.ngModel.currentValue);
+    }
+
+    if (changes.ngRequired) {
+      this.validate();
+    }
+  }
+
+  onExternalOptionsChange(newOptions) {
+    this.internalOptions = convertModelToOptions(this.ngModel, newOptions);
```
gittymcgit, line 28: shouldn't it be this.$ngModel?

src/forms/checkbox-group/checkbox-group.spec.js
```diff
@@ -2,104 +2,106 @@
...
     });
   });
 
-  describe('interactions', function() {
-    it('should change the model when checkbox gets checked', function() {
-      expect(getCheckBoxValue(checkbox()[0])).toBeFalsy();
-      button()[0].click();
-      expect(getCheckBoxValue(checkbox()[0])).toBe(true);
-      expect($ngModel().$viewValue).toEqual(["1"]);
+  describe('when the model that is passed in changes', () => {
```
gittymcgit, line 55: I think it also makes sense to test the case when options change

StevePole: Good call 👍

src/forms/checkbox-group/checkbox-group.controller.js
StevePole: No, $ngModel represents the controller behind the ngModel directive. Whereas ngModel is the bindings and therefore the value.  What I'll do, is rename $ngModel to $ngModelController.

PR #310, opened 2020-07-26
Change live-capture blob upload to File upload

jonstieglitz:
## Context

[Uploads from web are failing](https://production-logs.tw.ee/app/logtrail#/?q=%22Extension%20check%20failed%20for%20blob%22&h=All&t=last%20month&i=services-*&_g=()) for live capture photos that are uploaded using [PersistAsync](https://github.com/transferwise/mitigator-service/blob/b3e62363106cf9c250798f7904079f337b2e3eea/src/main/java/com/transferwise/mitigator/api/form/component/v2/verification/IdDocumentWithFrontFaceLivePhotoAlternativeV2.java#L48)

https://transferwise.slack.com/archives/C2SGDKP4Y/p1596072668439100

## Changes

I haven't finalized my changes but I'll either change the blob to a File right after it is captured or I'll rename the blob to have a JPEG extension before uploading (this seems riskier). 

## Considerations

These changes are hard to test! They involve styleguide-components (which is consumed somewhere) and webapp. Any guidance on testing would be greatly appreciated.

### Before
Screen Shot 2020-07-26 at 12 24 31 PM
Screen Shot 2020-07-26 at 12 00 48 PM

## Checklist

### DIFF
--- a/src/forms/upload/camera-capture/camera-capture.controller.js
+++ b/src/forms/upload/camera-capture/camera-capture.controller.js
@@ -240,7 +240,7 @@ function createUploadCallback($ctrl) {
     $ctrl.showVideoPreview = false;
     $ctrl.closeVideoStream();
     $ctrl.$scope.$applyAsync(() => {
-      const file = new File([blob], 'image.jpeg', { type: 'image/jpeg' })
+      const file = new File([blob], 'image.jpeg', { type: 'image/jpeg' });
       $ctrl.onCapture({ file });
     });
   };
--- a/src/forms/upload/camera-capture/camera-capture.controller.js
+++ b/src/forms/upload/camera-capture/camera-capture.controller.js
@@ -240,7 +240,8 @@ function createUploadCallback($ctrl) {
     $ctrl.showVideoPreview = false;
     $ctrl.closeVideoStream();
     $ctrl.$scope.$applyAsync(() => {
-      $ctrl.onCapture({ file: blob });
+      const file = new File([blob], 'image.jpeg', { type: 'image/jpeg' })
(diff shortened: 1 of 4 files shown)


src/forms/upload/camera-capture/camera-capture.controller.js
```diff
@@ -240,7 +240,8 @@ function createUploadCallback($ctrl) {
     $ctrl.showVideoPreview = false;
     $ctrl.closeVideoStream();
     $ctrl.$scope.$applyAsync(() => {
-      $ctrl.onCapture({ file: blob });
+      const file = new File([blob], 'image.jpeg', { type: 'image/jpeg' })
```
jonstieglitz, line 243: I think this is the way to go

src/forms/upload/services/async-file-saver.service.js
```diff
@@ -10,6 +10,10 @@ class AsyncFileSaver {
     }
     const formData = new FormData();
     const key = httpOptions.param || fieldName;
+
+    if (file[0].name === 'blob') {
+      formData.append(key, file, fieldName + file[0].ext);
+    }
     formData.append(key, file);
```
jonstieglitz, line 17: why would a blob have an extension?

tiborczvikli: why do not use else clause

jonstieglitz: you're right

jonstieglitz: but only one of these two fixes should be implemented, either the rename here or at capture

PR #317, opened 2020-08-06
[IDWL-] Change upload from live capture to add the file extension if its not exist

tiborczvikli:
## Context
Uploads from web are failing for live capture photos that are uploaded using `PersistAsync`

https://transferwise.slack.com/archives/C2SGDKP4Y/p1596072668439100

## Changes
As you can see at [processing-card.controller.js:54](https://github.com/transferwise/styleguide-components/pull/317/files#diff-edaa2d37a284510d3112b18de78ce543R54) we do compute the Base64 data URL of the `Blob` we get, but we are not using it, since the `mitigator-service` works with `MultipartFile`s instead of a `String` on that endpoint which would be risky to change.

Because we append a `Blob` instance to the form data without specifying the `fileName` argument, [it will default to the `blob` string](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) which is an invalid file name from the `mitigator-service` perspective.

To be backward compatible - if there is any other usages of the `AsyncFileSaver` service -, but still generic, we decided to try deducting the file extension from the already computed Base64 data URL, but if we fail to do so it just works as before.

## Considerations
Now the live image capture and upload with persist async to `mitigator-service` to `webapp` is failing. This change is a fix for that.
This is the cause of the failure: https://github.com/transferwise/webapp/blob/c981e42067d1321f4cac6ed463b36444b729b553/tw-plugins/tw-common/grails-app/services/com/transferwise/fx/user/document/FileUploadService.groovy#L96

### Before

image
image

## Checklist

### DIFF
--- a/src/forms/upload/processing-card/processing-card.controller.js
+++ b/src/forms/upload/processing-card/processing-card.controller.js
@@ -51,7 +51,7 @@ class Controller {
     if (this.httpOptions) {
       // Post file now
       this.asyncFileRead(file)
-        .then(dataUrl => this.asyncFileSave(file, this.parseFileName(dataUrl))
+        .then(dataUrl => this.asyncFileSave(file, Controller.parseFileName(dataUrl))
           .then(response => asyncSuccess(response, dataUrl, this)))
         .catch(error => asyncFailure(error, null, this));
     } else {
@@ -62,7 +62,7 @@ class Controller {
     }
   }
 
-  parseFileName(dataUrl) {
+  static parseFileName(dataUrl) {
     const match = dataUrl.match(/^data:[^/]+\/([^;]+);base64,/);
 
     return match ? `file.${match[1]}` : undefined;
--- a/src/forms/upload/processing-card/processing-card.controller.js
+++ b/src/forms/upload/processing-card/processing-card.controller.js
@@ -51,7 +51,7 @@ class Controller {
     if (this.httpOptions) {
       // Post file now
       this.asyncFileRead(file)
-        .then(dataUrl => this.asyncFileSave(file, dataUrl)
+        .then(dataUrl => this.asyncFileSave(file, this.parseFileName(dataUrl))
           .then(response => asyncSuccess(response, dataUrl, this)))
         .catch(error => asyncFailure(error, null, this));
     } else {
@@ -60,6 +60,12 @@ class Controller {
         .then(dataUrl => asyncSuccess(null, dataUrl, this))
         .catch(error => asyncFailure(error, null, this));
     }
+  }
+
+  parseFileName(dataUrl) {
+    const match = dataUrl.match(/^data:[^/]+\/([^;]+);base64,/);
+
+    return match ? `file.${match[1]}` : undefined;
   }
 
   asyncFileSave(file, dataUrl) {
--- a/src/forms/upload/processing-card/processing-card.controller.js
+++ b/src/forms/upload/processing-card/processing-card.controller.js
@@ -68,9 +68,9 @@ class Controller {
     return match ? `file.${match[1]}` : undefined;
   }
 
-  asyncFileSave(file, dataUrl) {
+  asyncFileSave(file, fileName) {
     const httpOptions = this.AsyncTasksConfig.extendHttpOptions(this.httpOptions);
-    return this.AsyncFileSaver.save(httpOptions.param || this.name, file, httpOptions, dataUrl);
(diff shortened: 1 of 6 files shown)


src/forms/upload/services/async-file-saver.service.js
```diff
@@ -10,7 +10,7 @@ class AsyncFileSaver {
     }
     const formData = new FormData();
     const key = httpOptions.param || fieldName;
-    formData.append(key, file);
+    formData.append(key, file, file[0].name === 'blob' ? 'blob.jpeg' : undefined);
```
StevePole, line 13: Why would the file have a name 'blob'?

StevePole:
I could be wrong about this, but I wonder if the problem is that the we allow the browser to choose the content type here:
https://github.com/transferwise/styleguide-components/pull/317/files#diff-98f9929680b3933767495af024a25811R36

Because we’re capturing from the screen, the file may be lacking the information the browser needs to choose the right content type.  I wonder if by changing the name you’re giving the browser the right information.  If that’s the case, the better thing to do would be to add this information somehow during capture of the image, rather than based on matching an arbitrary file name at the point of upload

src/forms/upload/services/async-file-saver.service.js
tiborczvikli: This is the file name by default and also do not contains the extension. The extension is required to be able to upload apparently. We recently tested that. I added logs on staging to webapp where its checked and the MIME type info is missing.
image

tiborczvikli: I created a request based on the original and added the file just renamed the filename and it worked well
image

tiborczvikli: Th only difference was the file name
Screen Shot 2020-08-03 at 12 24 45 PM (1)

tiborczvikli:
Sounds cool if we could achieve the same with that you mentioned ☝️ Could you share the link where it should be? Because right now its pointing to this PR @StevePole

StevePole:
Could you try renaming the file here instead: 
https://github.com/transferwise/styleguide-components/blob/2348166a5c329691892b09c90f5ffecd74824ccf/src/forms/upload/camera-capture/camera-capture.controller.js#L243

I don't like relying on the fact that the browser happened to call the blob 'blob' that seems like the kind of thing that may not be universal.

tiborczvikli:
> Could you try renaming the file here instead:
> https://github.com/transferwise/styleguide-components/blob/2348166a5c329691892b09c90f5ffecd74824ccf/src/forms/upload/camera-capture/camera-capture.controller.js#L243
> 
> I don't like relying on the fact that the browser happened to call the blob 'blob' that seems like the kind of thing that may not be universal.

Yes of course, I tried that solution too. I found out just one solution for that is 

      const file = new File([blob], 'image.jpeg', { type: 'image/jpeg' });
      $ctrl.onCapture({ file });

The new File is not supported in Internet Explorer AFAIK. I'm not sure we need to support that or there is another solution. @StevePole

StevePole:
What about if you try:
```
blob.name = 'image.jpg'
$ctrl.onCapture({ file: blob });
```
From what I see, that should be semantically equivalent to what you're already doing.  You check to see if that blob (which is now treated as a file) has the name 'blob' and rename it if not.  So this should be the same, just in a more appropriate spot where you know it's fine to rename the file.  It might be a read only property that's my only concern.

src/forms/upload/services/async-file-saver.service.js
sandorr: This change has been overwritten, please comment on the new implementation.

src/forms/upload/processing-card/processing-card.controller.js
```diff
@@ -62,9 +62,15 @@ class Controller {
...
   }
 
-  asyncFileSave(file) {
+  parseFileName(dataUrl) {
+    const match = dataUrl.match(/^data:[^/]+\/([^;]+);base64,/);
+
+    return match ? `file.${match[1]}` : undefined;
+  }
+
+  asyncFileSave(file, dataUrl) {
```
StevePole, line 71: Shouldn't `dataUrl` be `fileName`?

sandorr: I don't think so, because the input is the `dataUrl` and the output is the `fileName`

StevePole: Don't think so... It's already been converted to a filename on line 54.  I'm happy with everything other than this one  🙏

PR #322, opened 2020-08-21
Allow custom validation messages with defaults as fallback

lucasogura-tw:
## Context
Custom validation messages can be set for just particular fields. Previously you need to set it for all of the validations or none

For example if we set: 
```
"validationMessages": {
                              "pattern": "wrong format"
                          },
```

Missing validation message for `required` will fallback to default.

## Changes
Default values combine with custom validation messages

## Considerations
Added this after adding validation messages support to df v2 https://github.com/transferwise/dynamic-interfaces/commit/9963af0a5ce81801d82ccd484ff15ff65a0202cc

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v7.0.4
+## Fallback to default for validation message for individual fields
+Allows setting particular custom validation messages and the rest will fallback to default
+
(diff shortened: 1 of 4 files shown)


StevePole (approved):
👍

PR #324, opened 2020-09-02
fix: support V2 hidden fields in requirements service

StevePole:
## Context
The requirements service, which maps V2 dynamic form specifications to V3, was missing support for hidden field types.  This became apparent when we started to remove 'undocumented' fields, as we were removing values from hidden fields.

We were also removing data from select fields, because our 'model cleaning' code expects JSON schema style 'oneOf' selects, not 'values' arrays. 

## Changes
We now correctly map hidden fields to hidden string schemas.  We now add 'oneOf' arrays next to our 'values' array to enable validation to work correctly

## Considerations
n/a

## Original Pull Request (for version bumps)
n/a

## Screenshots/GIFs
n/a

### Before
n/a

### After
n/a

## Checklist


felixxiatransferwise (approved):
👍

PR #326, opened 2020-09-07
Fix values fields mapping for checkbox-group component

gittymcgit:
## Changes

Adding prepLegacyValues on checkbox group to map key/name to value/label

## Checklist

### DIFF
--- a/src/forms/fieldset/fieldset.demo.js
+++ b/src/forms/fieldset/fieldset.demo.js
@@ -272,8 +272,8 @@ function fieldsetDocsController() {
         items: {
           enum: [1, 2],
           values: [
-            { key: 1, name: 'One' },
-            { key: 2, name: 'Two' }
+            { value: 1, label: 'One' },
+            { value: 2, label: 'Two' }
           ]
         }
       },
--- a/src/forms/fieldset/fieldset.demo.js
+++ b/src/forms/fieldset/fieldset.demo.js
@@ -272,8 +272,8 @@ function fieldsetDocsController() {
         items: {
           enum: [1, 2],
           values: [
-            { value: 1, label: 'One' },
-            { value: 2, label: 'Two' }
+            { key: 1, name: 'One' },
+            { key: 2, name: 'Two' }
           ]
         }
       },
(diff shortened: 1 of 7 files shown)


src/forms/fieldset/fieldset.demo.js
```diff
@@ -272,8 +272,8 @@ function fieldsetDocsController() {
         items: {
           enum: [1, 2],
           values: [
-            { value: 1, label: 'One' },
-            { value: 2, label: 'Two' }
+            { key: 1, name: 'One' },
+            { key: 2, name: 'Two' }
```
oleksandrbylim, line 276: btw, this looks like contract change, was it changed in some previous commits?
is there same fix or reflection in dynamic-interfaces lib?

gittymcgit: This reflects current contract in dynamic interfaces
Requirements service should transform key and name to value and label when preparing the field

StevePole: I would prefer the fieldset demo and tests use the latest iteration of the spec.  We should write tests for the requirements service to make sure it correctly translates the older format to the new

gittymcgit: I see what you mean. Removed that change, added a test for requirements service to test that they're mapped

oleksandrbylim (approved):

src/services/requirements/requirements.spec.js
```diff
@@ -73,6 +73,25 @@ describe('Requirements Service', function() {
...
+        checkboxGroup: { type: "select", control: "select", selectType: "CHECKBOX",  valuesAllowed: [{ key: 1, name: "One"}] }
+      };
+
+      var current = {
+        checkboxGroup: {
+          items: {
+            values: [{ value: 1, label: "One"}]
+          },
+          control: "select",
+          selectType: "CHECKBOX",
```
StevePole, line 87: I don't think this is correct, selectType is not part of V3 spec, If you also implement my other suggestion, I think I would expect it to be:
```
{
  type: "array",
  items: {
    values: [{ value: 1, label: "One"}], 
    oneOf: [{ const: 1, title: "One" }]
  },
  control: "checkbox-group"
}
```
At some stage we may deprecate 'values', and fully port things to oneOf (that's what we're doing in the React version), for now oneOf lets us correctly validate.

StevePole (approved):
👍

PR #340, opened 2020-10-16
Fix icon problems related to live upload component

HaoYuanTW:
## Context
2 problems here
1. TW icon does not scale to button size the way glyphicon does making it off center and ugly
2. "Tick" icon does not exist

Which makes our camera upload component look like this
Screenshot 2020-10-16 at 16 29 37

## Changes
Fix the button problems. It seems our icon assets are not made to scale nicely. I understand there is another way to do it with svg viewbox attribute, but that solution wont wont apply here. 

Updated spec is to use a 56px button with a 24px icon as advised by design team. 

### After
Screenshot 2020-10-16 at 16 59 34

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v8.0.3
+## Fix cancel and confirm buttons for camera upload component
+
 # v8.0.2
 ## Force a digest when a file input value changes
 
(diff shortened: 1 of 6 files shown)


sneakertack (approved):

StevePole (approved):
Please bump to 8.0.3 and add changelog entry

PR #342, opened 2020-11-06
feat: microapps media API support

ekwonye-richard:
## Context

There is a new integration with google where we need to render transferwise.com inside https://microapps.google.com domain. There will be an android native application with a webview which will open transferwise.com inside.

We need to use Media API exposed by microapps for file upload, hence we need to make this the preferred option when the upload component is rendered inside the supported domain.

## Changes

Use Media API for file upload when the upload component is rendered inside the supported domain.

## Checklist

### DIFF
--- a/src/forms/upload/capture-card/capture-card.controller.js
+++ b/src/forms/upload/capture-card/capture-card.controller.js
@@ -1,8 +1,17 @@
 class Controller {
-  constructor($element) {
+  constructor($window, $element, MediaApiService) {
+    this.$window = $window;
     this.$element = $element;
     this.showLiveCaptureScreen = false;
     this.cameraFailed = false;
+    this.MediaApiService = MediaApiService;
+    this.checkMediaUploadSupport = this.checkMediaUploadSupport.bind(this);
+
+    if (this.$window.document.readyState === 'complete') {
+      this.checkMediaUploadSupport();
+    } else {
+      this.$window.addEventListener('load', this.checkMediaUploadSupport);
+    }
   }
 
   $onChanges(changes) {
@@ -10,8 +19,16 @@ class Controller {
       if ((changes.icon || {}).currentValue) {
         this.viewIcon = changes.icon.currentValue;
       } else {
-        this.viewIcon = (changes.isLiveCameraUpload || {}).currentValue ? 'camera' : 'upload';
+        this.viewIcon = (this.isMediaUpload || (changes.isLiveCameraUpload || {}).currentValue) ? 'camera' : 'upload';
       }
+    }
+  }
+
+  checkMediaUploadSupport() {
+    this.isMediaUpload = this.MediaApiService.hasMediaUploadSupport();
+
+    if (this.isMediaUpload) {
+      this.viewIcon = 'camera';
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/forms/upload/upload.controller.js
```diff
@@ -1,15 +1,26 @@
...
+    MediaApiService
   ) {
+    this.$window = $window;
     this.$element = $element;
     this.$attrs = $attrs;
     this.FileValidationService = FileValidationService;
+    this.MediaApiService = MediaApiService;
 
     this.isProcessing = false;
+    this.checkMediaUploadSupport = this.checkMediaUploadSupport.bind(this);
```
StevePole, line 17: I feel like we could encapsulate all of this logic within the capture card.  It's really only a concern for the component responsible for capturing the file, and it doesn't appear to have any impact on this controller/template, except for passing down the state to the capture card.

src/forms/upload/services/media-api.service.js
```diff
@@ -0,0 +1,61 @@
+import angular from 'angular';
+
+class MediaApiService {
```
StevePole, line 3: I like how you encapsulated this stuff in the service 👍

src/forms/upload/capture-card/capture-card.controller.js
```diff
@@ -38,8 +51,21 @@ class Controller {
     const file = files[0];
     this.onFileCapture({ file });
   }
+
+  captureFromMedia() {
```
StevePole, line 55: Perhaps this logic could be encapsulated in the service as well?

src/forms/upload/upload.controller.js
ekwonye-richard: Yes, I agree with you on this. I'll move this to card capture

src/forms/upload/capture-card/capture-card.controller.js
ekwonye-richard: Yes, we can. I'll move it too.

src/forms/upload/capture-card/capture-card.controller.js
```diff
@@ -1,12 +1,21 @@
...
+    if (this.$window.document.readyState === 'complete') {
+      this.checkMediaUploadSupport();
+    } else {
+      this.$window.addEventListener('load', this.checkMediaUploadSupport);
+    }
   }
 
   $onChanges(changes) {
-    if (changes.icon || changes.isLiveCameraUpload) {
+    if ((changes.icon || changes.isLiveCameraUpload) && !this.isMediaUpload) {
```
StevePole, line 18: I don't see a problem, but interested why this needs to change?

src/forms/upload/services/media-api.service.js
```diff
@@ -0,0 +1,76 @@
...
```
StevePole, line 3: Could we perhaps add a doc block to explain what this service is all about.  I worry that someone coming to this fresh would have no context for this.

src/forms/upload/capture-card/capture-card.controller.js
ekwonye-richard: If media upload is supported it's good we keep the default icon as camera.

ekwonye-richard: On a second thought, I should make this check only when icon isn't changed.

src/forms/upload/services/media-api.service.js
ekwonye-richard: I will add this

StevePole (approved):
Nice tests 👍 

There may be a little repetition with the upload tests.  I think previously we were treating the whole upload as the testable 'unit', hence the card itself didn't have tests.  Though I'm sure what you've added is more thorough anyway.

PR #348, opened 2020-12-03
Throw an error if we’re in the Spot environment but microapps API is not available

ekwonye-richard:
## Context

We previously checked if microapps is in window and if it isn't we fall back to HTML file input or the camera component.
Spot app does not support html file input and does not throw errors when trying using it.
We will always default to use the Media API when within a spot host url which will then throw an error if microapps is not available in window.

## Checklist

### DIFF
--- a/src/forms/upload/capture-card/card-capture.spec.js
+++ b/src/forms/upload/capture-card/card-capture.spec.js
@@ -86,79 +86,81 @@ describe('given a Card Capture component', () => {
   })
 
   describe('when media upload is supported', () => {
-    let deferred;
+    describe('when microapps is in window', () => {
+      let deferred;
 
-    beforeEach(() => {
-      deferred = $q.defer();
+      beforeEach(() => {
+        deferred = $q.defer();
 
-      jest.spyOn(MediaApiService, 'hasMediaUploadSupport').mockReturnValue(true);
-      jest.spyOn(MediaApiService, 'captureFromMedia').mockReturnValue(deferred.promise);
-    })
+        jest.spyOn(MediaApiService, 'hasMediaUploadSupport').mockReturnValue(true);
+        jest.spyOn(MediaApiService, 'captureFromMedia').mockReturnValue(deferred.promise);
+      })
 
-    it('should render a camera button', () => {
-      expect(findCameraButton()).toBeTruthy();
-    })
+      it('should render a camera button', () => {
+        expect(findCameraButton()).toBeTruthy();
+      })
 
-    describe('when isLiveCameraUpload is true', () => {
-      describe('when camera button is clicked', () => {
+      describe('when isLiveCameraUpload is true and camera button is clicked', () => {
         beforeEach(() => {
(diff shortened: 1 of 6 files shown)


src/forms/upload/capture-card/card-capture.spec.js
```diff
@@ -86,79 +86,87 @@ describe('given a Card Capture component', () => {
...
+      describe('when isLiveCameraUpload is true', () => {
+        describe('when camera button is clicked', () => {
+          beforeEach(() => {
+            $scope.isLiveCameraUpload = true;
 
-          compileComponent();
+            compileComponent();
+          })
+
+          it('should call captureFromMedia from MediaApiService with true when camera button is clicked', () => {
```
StevePole, line 111: Don't think you need to say when clicked here, or actually do the click as the preceding describe/before each does it?

src/forms/upload/capture-card/card-capture.spec.js
```diff
@@ -86,79 +86,87 @@ describe('given a Card Capture component', () => {
...
+        describe('when camera button is clicked', () => {
+          beforeEach(() => {
+            $scope.isLiveCameraUpload = false;
+
+            compileComponent();
+          })
+
+          it('should call captureFromMedia from MediaApiService with false when camera button is clicked', () => {
+            findCameraButton().click();
+            expect(MediaApiService.captureFromMedia).toHaveBeenCalledWith(false);
```
StevePole, line 128: Same, move the click into the before each

src/forms/upload/capture-card/card-capture.spec.js
ekwonye-richard: yes, good catch

src/forms/upload/capture-card/card-capture.spec.js
```diff
@@ -86,79 +86,85 @@ describe('given a Card Capture component', () => {
...
-        const file = { size: 124, type: 'image/jpeg'};
+          beforeEach(() => {
+            deferred.resolve(file);
+            $scope.$apply();
+          })
 
-        beforeEach(() => {
-          deferred.resolve(file);
-          $scope.$apply();
+          it('should call onFileCapture with ressolved file', () => {
```
StevePole, line 147: Typo

src/forms/upload/capture-card/card-capture.spec.js
```diff
@@ -86,79 +86,85 @@ describe('given a Card Capture component', () => {
...
-      describe('when camera button is clicked', () => {
-        beforeEach(() => {
-          $scope.isLiveCameraUpload = true;
+      it('should render a camera button', () => {
+        expect(findCameraButton()).toBeTruthy();
+      })
 
-          compileComponent();
+      describe('when isLiveCameraUpload is true', () => {
+        describe('when camera button is clicked', () => {
```
StevePole, line 104: You could combine these two describes, "when a and b"

StevePole (approved):
Couple minor optional things, but it's fine 👍

PR #352, opened 2021-01-27
VUX-547 - Disable drag and drop on non eligible cases

diego-transferwise:
## Context
The camera only feature of the upload component was not blocking customers to use the drag and drop behavior, so this PR is fixing it.

## Changes
https://transferwise.atlassian.net/browse/VUX-547

## Checklist


StevePole (approved):
LGTM

PR #355, opened 2021-02-18
breaking: remove requiredFields prop from Fieldset

StevePole:
## Context
Required status should come from the fields themselves, not the separate prop. There are no cases in our code base where we depend on this prop, so this should not a breaking change in practice.

## Changes
Removes the requiredFields prop from Fieldset

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# v9.0.0
+## Removes requiredFields prop from Fieldset
+Required status should come from the fields themselves, not the separate prop. There are no cases in our code base where we depend on this prop, so this should not a breaking change in practice.
+
 # v8.3.1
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# v9.0.0
```
mSobhy90, line 1: 💅 If we're not using it should we make a major version bump?

mSobhy90 (approved):
LGTM 🚀 
It's interesting we never used that property on mobile. 🤷

CHANGELOG.md
StevePole: It's technically breaking, and this library is open source, so I thought better to stick to semver.

PR #363, opened 2021-06-11
fix(Field): handle uppercase field types as well

pterblgh:
## Context

`twField` component incorrectly mapped upper case field type names (`number` vs `NUMBER`) and this resulted in fields having `text` control type - while they should have been something different.

This caused validation issues. Let's take for example field type `NUMBER` - this is the value that's being returned by Dynamic Flows V2 (DF V2) field type value on the network. Example network response for such field:
```
            {
              '@class': 'ee.tw.dynamic.interfaces.structure.field.NumberField',
              key: 'transferId',
              name: 'Transfer number',
              description: null,
              width: 'lg',
              type: 'NUMBER',
              required: true,
              hidden: false,
              placeholder: '12345678',
              helpText: null,
              columns: null,
              validationRegexp: '\\d+',
              validationMessages: null,
              refreshRequirementsOnChange: false,
              min: 0,
              max: 2147483647,
              example: null,
            },
```

In the client side code data flows from `twField` to `twFormControl`, something like this:

```html
<tw-field>
  <tw-form-control type="{{ $ctrl.control | lowercase }}" />
</tw-field>
```

When field data arrives to `twField`, the component transforms the `field` object according to legacy rules - this is done in `RequirementsService.prepField`. This method converts the `field.type` value to a lowercase string. 

At line 14 `this.field.type` would be `"number"`.  So far so good.

https://github.com/transferwise/styleguide-components/blob/d62284fcfcd5a33d63d02b1b0ca3c22642e59967/src/forms/field/field.controller.js#L11-L16

However at line 15 we pass down the original `field` POJO - not the one that we've just transformed. So in this case `field.type` is `"NUMBER"`.

The value returned by `RequirementsService.getControlType` is being passed down to the `twFormControl` component. If you check the current diff, you can see that this method expects that `field.type` is a lowercase string. Remember, we call `getControlType` with the original `field` object and not the one where we already transformed the `field.type` value to lowercase.

So at the end when this function gets called with `field.type` having a value of `"NUMBER"` for example, according to the current implementation this function returns `text` as a fallback in the switch-case. This is problematic.

In this particular `"NUMBER"` example this behaviour breaks the validation of the `field` and the component won't emit a "valid model" event. Due to this the `twFieldset` component remains "invalid", thus the `twForm` is not valid so the current DF alternative form is not submittable so customers get stuck submitting this step.

## Changes

This PR makes sure that `RequirementsService.getControlType` transforms `field.type` to a lowercase string just like `RequirementsService.prepField` does.

## Considerations

We might change the argument of `RequirementsService.getControlType` at line 15 to `this.field` - I am just not sure whether that would have any side-effects.

https://github.com/transferwise/styleguide-components/blob/d62284fcfcd5a33d63d02b1b0ca3c22642e59967/src/forms/field/field.controller.js#L11-L16

## Checklist

### DIFF
--- a/src/forms/field/field.spec.js
+++ b/src/forms/field/field.spec.js
@@ -129,25 +129,19 @@ describe('Field', function() {
     });
   });
 
-  describe('when given type:number', function() {
-    beforeEach(function() {
-      $scope.field = { type: "number" };
+  describe.each([
+    ["number", "number"],
+    ["NUMBER", "number"],
+    ["boolean", "checkbox"],
+    ["BOOLEAN", "checkbox"],
+  ])("when given type:%s", (type, control) => {
+    beforeEach(function () {
+      $scope.field = { type };
       element = getCompiledDirectiveElement();
     });
 
-    it('should ask the form control to render a number input', function() {
-      expect(FormControl.bindings.type).toBe('number');
-    });
-  });
-
-  describe('when given type:boolean', function() {
-    beforeEach(function() {
-      $scope.field = { type: "boolean" };
-      element = getCompiledDirectiveElement();
-    });
-
-    it('should ask the form control to render a telephone input', function() {
-      expect(FormControl.bindings.type).toBe('checkbox');
+    it("should ask the form control to render a number input", function () {
+      expect(FormControl.bindings.type).toBe(control);
     });
   });
 
--- a/src/forms/field/field.spec.js
+++ b/src/forms/field/field.spec.js
@@ -130,17 +130,19 @@ describe('Field', function() {
   });
 
   describe.each([
-    ["number", "number"],
-    ["NUMBER", "number"],
-    ["boolean", "checkbox"],
-    ["BOOLEAN", "checkbox"],
+    ['number', 'number'],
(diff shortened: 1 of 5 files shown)


src/forms/field/field.spec.js
```diff
@@ -129,25 +129,21 @@ describe('Field', function() {
...
+    ['BOOLEAN', 'checkbox'],
+  ])("when given type:%s", (type, control) => {
+    beforeEach(function () {
+      $scope.field = { type };
       element = getCompiledDirectiveElement();
     });
 
-    it('should ask the form control to render a telephone input', function() {
-      expect(FormControl.bindings.type).toBe('checkbox');
+    it('should ask the form control to render a number input', function () {
```
StevePole, line 145: Some of these are rendering a checkbox...

StevePole (approved):
Looks good apart from that tiny test naming issue.

src/forms/field/field.spec.js
pterblgh: Nice catch, fixed it! 👍

PR #373, opened 2022-01-24
new disabled state approach

dozortsev:
## Context

Neptune Design System got new disabled states.

## Changes

- update styles foundation (neptune-css), this does 99% of work
- also, needed to change location of `disabled` class in a few components because otherwise selectors like `:hover:not(.disabled, disabled)` won't work and component would change some of it's colors on hover or active events

### Before

**http://transferwise.github.io/styleguide-components/**

### After

**https://transferwise.github.io/styleguide-components/branch/new-disabled-state-approach/index.html**

## Checklist


src/forms/form-control/form-control.html
```diff
@@ -2,10 +2,10 @@
   <div
     ng-switch-when="radio"
     class="radio"
-    ng-class="{disabled: $ctrl.ngDisabled, 'radio-lg': option.secondary}"
+    ng-class="{'radio-lg': option.secondary}"
     ng-repeat="option in $ctrl.options"
   >
-    <label>
+    <label ng-class="{'disabled': $ctrl.ngDisabled}">
```
dozortsev, line 8: needed to change location of `disabled` class because otherwise `:hover:not(.disabled, disabled)` won't work and component would change it's border colors on hover

nabaroa (approved):

JulianG-TW (approved):

PR #404, opened 2023-01-18
github actions

jonstieglitz:
## Checklist


src/forms/select/select.less
```diff
@@ -0,0 +1,3 @@
+.form-group .input-group .tw-select .btn.dropdown-toggle {
+  border-radius: 10px;
+}
```
jonstieglitz, line 3: Fix for select double border bug

